PR #8, opened 2018-02-23
Add function to decrypt with a given set of keys instead of using the key cache dir

elliott-davis:
Signed-off-by: Elliott Davis <elliott@excellent.io>

### DIFF
--- a/components/core/src/crypto/keys/box_key_pair.rs
+++ b/components/core/src/crypto/keys/box_key_pair.rs
@@ -28,6 +28,14 @@ use super::{get_key_revisions, mk_key_filename, mk_revision_string, parse_name_w
 use super::super::{ANONYMOUS_BOX_FORMAT_VERSION, BOX_FORMAT_VERSION, PUBLIC_BOX_KEY_VERSION,
                    PUBLIC_KEY_SUFFIX, SECRET_BOX_KEY_SUFFIX, SECRET_BOX_KEY_VERSION};
 
+#[derive(Debug)]
+pub struct BoxSecret<'a> {
+    pub sender: &'a str,
+    pub ciphertext: Vec<u8>,
+    pub receiver: Option<&'a str>,
+    pub nonce: Option<Nonce>,
+}
+
 pub type BoxKeyPair = KeyPair<BoxPublicKey, BoxSecretKey>;
 
 impl BoxKeyPair {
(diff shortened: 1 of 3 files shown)


thesentinels:

Thank you for contributing!

components/core/src/crypto/keys/box_key_pair.rs
```diff
@@ -238,90 +238,131 @@ impl BoxKeyPair {
...
-    /// Key names are embedded in the message payload which must
-    /// be present while decrypting.
-    pub fn decrypt<P>(payload: &[u8], cache_key_path: P) -> Result<Vec<u8>>
-    where
-        P: AsRef<Path>,
-    {
-        debug!("Decrypt key path = {}", cache_key_path.as_ref().display());
-        let mut lines = str::from_utf8(payload)?.lines();
-        let version = match lines.next() {
+    pub fn box_key_version(version: Option<&str>) -> Result<&str> {
```
chefsalim: should this be named box_key_format_version to differentiate better between key and box format versions?

components/core/src/crypto/keys/box_key_pair.rs
```diff
@@ -238,90 +238,131 @@ impl BoxKeyPair {
...
                 if val != BOX_FORMAT_VERSION && val != ANONYMOUS_BOX_FORMAT_VERSION {
                     return Err(Error::CryptoError(format!("Unsupported version: {}", val)));
                 };
-                val
+                Ok(val)
             }
             None => {
-                return Err(Error::CryptoError(
+                Err(Error::CryptoError(
                     "Corrupt payload, can't read file version".to_string(),
```
chefsalim: Update error to remove reference to 'file' (assuming this is used for both in-memory and file checks)?

components/core/src/crypto/keys/box_key_pair.rs
```diff
@@ -238,90 +238,131 @@ impl BoxKeyPair {
...
-        let ciphertext = match lines.next() {
+            None => {
+                Err(Error::CryptoError(
+                    "Corrupt payload, can't read nonce".to_string(),
+                ))
+            }
+        }
+    }
+
+    pub fn box_key_cyphertext(cyphertext: Option<&str>) -> Result<Vec<u8>> {
```
chefsalim: Rename to 'ciphertext' to be consist with the spelling elsewhere?

elliott-davis: :rofl: spelling, how does it work! Thanks

components/core/src/crypto/keys/box_key_pair.rs
elliott-davis: The hardest part of this PR has been the function names.

components/core/src/crypto/keys/box_key_pair.rs
```diff
@@ -238,90 +238,131 @@ impl BoxKeyPair {
...
+            Some(recv) => {
+                Self::decrypt_box(ciphertext, &nonce.unwrap(), self.public()?, recv.secret()?)
             }
+            None => Self::decrypt_anonymous_box(ciphertext, self.public()?, self.secret()?),
+        }
+    }
+
+    // Return the metadata and encrypted text from a secret payload.
+    // This is useful for services consuming an encrypted payload and need to decrypt it without having keys on disk
+    pub fn secret_metadata(payload: &[u8]) -> Result<(&str, Vec<u8>, Option<&str>, Option<Nonce>)> {
```
chefsalim: This almost feels like it should be it's own entity instead of a helper in BoxKeyPair... maybe we need to tease out the concept of a "Box" as it's own struct.

components/core/src/crypto/keys/mod.rs
```diff
@@ -391,6 +393,108 @@ where
...
+/// }
+/// ```
+///
+/// # Errors
+///
+/// * If there is a key version mismatch
+/// * If the key version is missing
+/// * If the key name with revision is missing
+/// * If the key value (the Bas64 payload) is missing
+pub fn parse_key_str(content: &str) -> Result<(PairType, String, String)> {
```
chefsalim: Is the notion that we expose this directly to external consumers?  If so, I'm not sure that is the right model...maybe we should look at something like a FromStr for the KeyPair struct?

elliott-davis: we were already exposing it to external customers in SigKeyPair but I think I agree (contingent on the effort required to make it work across the entire code base)

components/core/src/crypto/keys/box_key_pair.rs
elliott-davis: So you are suggesting something like:
```
struct BoxSecret {
    sender: &str,
    ciphertext: Vec<u8>,
    receiver: Option<&str>,
    nonce: Option<Nonce>,
}
```

components/core/src/crypto/keys/mod.rs
elliott-davis: Okay after looking at this I think it's a pretty big lift. There are several functions in there that could and should be replaces with From/ToString impls. I feel like that should go in a separate PR, which I am also happy to take on, but as a separate change.

fnichol: I'm going to bet (having helped with some of this code) that we hadn't yet fallen in love with the `FromStr` trait.

elliott-davis:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/core/builds/347763435).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/core/builds/347763435).

PR #9, opened 2018-02-27
Set up a SIGCHLD handler on Linux

christophermaier:
This will allow us to reap zombies when running the Launcher as PID 1. 
Without this, zombie processes would build up on a system, eventually 
filling up the process table. This would be a Bad Thing (TM).

While explicitly ignoring `SIGCHLD` with `SIG_IGN` would eliminate the
zombie problem, we would run into problems when the child process is
the Supervisor! We may need to restart it, so we have to actually
handle `SIGCHLD`.

Signed-off-by: Christopher Maier <cmaier@chef.io>


thesentinels:

Thank you for contributing!

PR #12, opened 2018-03-06
 Use a backslash for windows FS_ROOT and PKG_PATH

mwrock:
This ensures that the default `FS_ROOT` on windows ends in `\`. While almost all apps on windows handle forward slashes just as well as backslashes there are a few (like SQL Server 2016 (unlike 2017)) that may not understand forward slashes. This should fix those cases.

### DIFF
--- a/components/core/src/fs.rs
+++ b/components/core/src/fs.rs
@@ -428,7 +428,7 @@ fn fs_root_path() -> PathBuf {
     if cfg!(target_os = "windows") {
         match (henv::var(FS_ROOT_ENVVAR), henv::var(SYSTEMDRIVE_ENVVAR)) {
             (Ok(path), _) => PathBuf::from(path),
-            (Err(_), Ok(system_drive)) => PathBuf::from(format!("{}{}", system_drive, "/")),
+            (Err(_), Ok(system_drive)) => PathBuf::from(format!("{}{}", system_drive, "\\")),
             (Err(_), Err(_)) => {
                 unreachable!(
                     "Windows should always have a SYSTEMDRIVE \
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

mwrock:
related to https://github.com/habitat-sh/habitat/pull/4713

christophermaier:
@mwrock This looks like it adds the platform-specific separator to the root portion of the path; do we need to ensure that paths based on this are consistent in their use of the right separator? I recall in the past you mentioned that Windows should be able to handle either, so I think we may have some places where path separators might be hard-coded with `/`... if there are older apps that might not be able to cope, there may be places in the code that will trip them up.

mwrock:
Yeah windows predominately is OK with forward slashes. However sometimes (very rarely in my experience) an individual application might have some internal logic (like a regex path checker) that barfs on a forward slash. We came upon such a case with Sql Server 2016 and only occurs in a particular action (CREATE TABLE). Sql Server 2017 seems to have fixed it but the customer is bound to 2016.

This issue should be isolated to handlebar variable expansions performed by the supervisor. That would be all of the pkg.svc_* vars and the pkg.path var and the root var. Beyond these, its probably out of scope to try to make sure we use backslashes everywhere.

christophermaier:
@mwrock if you've confirmed that those places use the right path separator, I'm 👍 on this

mwrock:
yeah I tested a build and it came out legit yo.

christophermaier:
tenor-108183158

christophermaier:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/core/builds/352537186).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/core/builds/352537186).

PR #16, opened 2018-03-12
Exposes members from group struct via new fn

eeyun:
This PR is required to resolve https://github.com/habitat-sh/habitat/issues/4716
Exposes the members of the `group` struct from the user crate so that we can validate in the hab binary on studio entrance 

### DIFF
--- a/components/core/src/os/users/linux.rs
+++ b/components/core/src/os/users/linux.rs
@@ -15,7 +15,7 @@
 use std::path::PathBuf;
 
 use linux_users;
-use linux_users::os::unix::UserExt;
+use linux_users::os::unix::{GroupExt, UserExt};
 
 pub fn get_uid_by_name(owner: &str) -> Option<u32> {
     linux_users::get_user_by_name(owner).map(|u| u.uid())
@@ -23,6 +23,10 @@ pub fn get_uid_by_name(owner: &str) -> Option<u32> {
 
 pub fn get_gid_by_name(group: &str) -> Option<u32> {
     linux_users::get_group_by_name(&group.as_ref()).map(|g| g.gid())
+}
+
+pub fn get_members_by_groupname(group: &str) -> Option<Vec<String>> {
+    linux_users::get_group_by_name(&group.as_ref()).map(|g| g.members().to_vec())
 }
 
 pub fn get_current_username() -> Option<String> {


thesentinels:

Thank you for contributing!

components/core/src/os/users/linux.rs
```diff
@@ -15,6 +15,7 @@
 use std::path::PathBuf;
 
 use linux_users;
+use linux_users::os::unix::GroupExt;
 use linux_users::os::unix::UserExt;
```
raskchanky: You can combine these two lines together, like so:

```rust
use linux_users::os::unix::{GroupExt, UserExt};
```

eeyun: Ah yep, Indeed, good  catch.

components/core/src/os/users/linux.rs
```diff
@@ -25,6 +26,13 @@ pub fn get_gid_by_name(group: &str) -> Option<u32> {
     linux_users::get_group_by_name(&group.as_ref()).map(|g| g.gid())
 }
 
+pub fn get_members_by_groupname(group: &str) -> Vec<String> {
+    match linux_users::get_group_by_name(group) {
+        Some(groupmembers) => groupmembers.members().to_vec(),
+        None => Vec::new(),
+    }
+}
+
```
raskchanky: In keeping with the spirit of the `users` crate, if you changed this function to return an `Option`, you could simplify it down to:

```rust
pub fn get_members_by_groupname(group: &str) -> Option<Vec<String>> {
    linux_users::get_group_by_name(&group.as_ref()).map(|g| g.members().to_vec())
}
```

Though I don't actually see where the `members()` function is defined 😁

raskchanky: I say "in keeping with the spirit of the `users` crate" because it seems like that crate chooses to return `None` if the group doesn't exist, which also seems to make sense here (I think).

eeyun: So members is a trait on the GroupExt struct. I think this is probably a totally reasonable concession.

raskchanky:
@thesentinels approve

thesentinels:

thesentinels:
:neckbeard: Travis CI has [started testing this PR](https://travis-ci.org/habitat-sh/core/builds/353053139).

thesentinels:
:sparkling_heart: Travis CI [reports this PR passed](https://travis-ci.org/habitat-sh/core/builds/353053139).

PR #19, opened 2018-04-10
[core] Fix 2 issues with interpretation and use of `PATH` metafiles

fnichol:
There are 2 fixes in this changeset, both of which impact the way the `PATH` metafile is used and how older Habitat packages are treated by current tooling like the `hab` CLI, exporters, and the Supervisor.

## [core] Restore impl for PackageInstall#paths() to use `PATH` metafile.

This change restores the previous implementation of the
`PackageInstall#paths()` method to return path entries directly from the
`PATH` metafile and to not use the `RUNTIME_ENVIRONMENT` metafile.

As referenced in habitat-sh/builder#240, the semantic meaning of the
`PATH` metafile is precise and distinct from the newer
`RUNTIME_ENVIRONMENT` metafile. The path entries in a `PATH` metafile
are a one-to-one mapping with the values in the `$pkg_bin_dirs` Plan
variable which created the package. That is, the path entries in `PATH`
should only refer to directories under the package's own
`$pkg_prefix`--not to any other package. There are certain activities
such as searching for a binary in a package, performing binlinking, etc.
which use the `PATH` metafile and addition entires from external
packages break these implementations.

All packages built with a build program after
habitat-sh/habitat@13344a6
which merged in habitat-sh/habitat#4067 (and released in Habitat 0.50.0)
may contain extra path entries in their `PATH` metafiles. To accommodate
these packages, an additional filtering step is performed in the
`#paths()` method to drop all path entries that fall outside of the
package's `$pkg_prefix`. Several net-new unit tests are added with this
change to ensure the correct behavior in future feature work.

## [core] Restore impl for PackageInstall#legacy_runtime_path().

This change restores a previous implementation of the
`PackageInstall#runtime_path()` method but renames it to
`PackageInstall#legacy_runtime_path()` to be used by older packages
which lack the `RUNTIME_ENVIRONMENT` metafile.

When verifying the runtime path ordering in the previous commit, the
calculated path ordering wasn't correct with respect to direct
dependencies vs. transitive dependencies.

The preserved reference implementation can be found in an older version
of the `habitat-sh/habitat` repository:

https://github.com/habitat-sh/habitat/blob/333b75d6234db0531cf4a5bdcb859f7d4adc2478/components/core/src/package/install.rs#L321-L350

In short, the path ordering needs to be:

1. All path entries from the package itself
2. All path entries (ordered as read from the `PATH` metafile) from each
direct dependency, ordered as read from the `DEPS` metafile
3. All path entries (ordered as read from the `PATH` metafile) from any
remaining transitive dependency, ordered as read from the `TDEPS`
metafile (which is lexically sorted at build time)

Ideally there shouldn't be any path entry duplications, but the ordering
is more important than any duplication. The corect ordering ensures that
the earlier entries in a Plan's `$pkg_deps` will be on `$PATH` first
(and therefore will "win" if another program of the same name comes from
another package also on `$PATH`).

A net-new unit test was added in this change to ensure the correct
behavior in future feature work.

### DIFF
--- a/components/core/src/package/install.rs
+++ b/components/core/src/package/install.rs
@@ -422,59 +422,42 @@ impl PackageInstall {
         Ok(deps)
     }
 
-    // Determine the complete path for this package, taking into
-    // account all its dependencies. Normally we would take this from
-    // `RUNTIME_ENVIRONMENT`, but for packages that were made before
-    // that was introduced, this is how we compute it.
-    fn full_legacy_path(&self) -> Result<Vec<PathBuf>> {
-        let mut full_path = Vec::new(); // TODO (CM): determine capacity first
+    /// Returns an ordered `Vec` of path entries which can be used to create a runtime `PATH` value
+    /// when an older package is missing a `RUNTIME_ENVIRONMENT` metafile.
+    ///
+    /// The path is constructed by taking all `PATH` metafile entries from the current package,
+    /// followed by entries from the *direct* dependencies first (in declared order), and then from
+    /// any remaining transitive dependencies last (in lexically sorted order). All entries are
+    /// present once in the order of their first appearance.
+    ///
+    /// Preserved reference implementation:
+    /// https://github.com/habitat-sh/habitat/blob/333b75d6234db0531cf4a5bdcb859f7d4adc2478/components/core/src/package/install.rs#L321-L350
+    fn legacy_runtime_path(&self) -> Result<Vec<PathBuf>> {
+        let mut paths = Vec::new();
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

fnichol:
I should note that I've already performed a survey of the codebase in `habitat-sh/habitat` and amazingly there's no code change required--every place we currently use `PackageInstall#paths()` is the correct call and meaning, and everywhere else we call `PacakgeInstall#runtime_environment()`.

fnichol:
Awesome, thanks!

PR #20, opened 2018-04-10
[core] Account for FS_ROOT in `PackageInstall#paths()` & refactor tests.

fnichol:
This change deals with an issue where the test suite was creating
incorrect data for `PATH` metafiles (they contained FS_ROOT path
prefixes) which caused the implementation of `PackageInstall#paths()` to
incorrectly pass when it should have failed. When looking at the testing
setups, a more real package installation would have better revealed the
bug so the test suite was refactored to create full install paths with a
temporary FS_ROOT, for example:
`/tmp/fs-root.Xxabc123efd/hab/pkg/acme/hotel/1.0.0/20180410160423` where
`/tmp/fs-root.Xxabc123efd` is the FS_ROOT.

### DIFF
--- a/components/core/src/package/install.rs
+++ b/components/core/src/package/install.rs
@@ -371,6 +371,9 @@ impl PackageInstall {
     pub fn paths(&self) -> Result<Vec<PathBuf>> {
         match self.read_metafile(MetaFile::Path) {
             Ok(body) => {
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

PR #31, opened 2018-05-10
Update http client timeout, add env

chefsalim:
We need to update the http client timeout and make it configurable in order to support uploading large files to the depot.

Signed-off-by: Salim Alam <salam@chef.io>

tenor-237610560

### DIFF
--- a/components/http-client/src/api_client.rs
+++ b/components/http-client/src/api_client.rs
@@ -33,7 +33,7 @@ use proxy::{proxy_unless_domain_exempted, ProxyInfo};
 use ssl;
 
 // Read and write TCP socket timeout for Hyper/HTTP client calls.
-const CLIENT_SOCKET_RW_TIMEOUT: u64 = 60;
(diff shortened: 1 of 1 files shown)


components/http-client/src/api_client.rs
```diff
@@ -33,7 +33,7 @@ use proxy::{proxy_unless_domain_exempted, ProxyInfo};
 use ssl;
 
 // Read and write TCP socket timeout for Hyper/HTTP client calls.
-const CLIENT_SOCKET_RW_TIMEOUT: u64 = 60;
+const CLIENT_SOCKET_RW_TIMEOUT: u64 = 120;
```
christophermaier: Might be nice to update the name of this constant to reflect the units involved (e.g. `CLIENT_SOCKET_RW_TIMEOUT_SEC`). That tends to make the code easier to read, and allows you to catch obvious mismatches.

chefsalim: Thanks, will update

PR #42, opened 2018-07-17
[core] Introduce a new `x86_64-linux-kernel2` `PackageTarget`.

fnichol:
Represents an **older** [Linux kernel]-based system from the 2.6.x
family running on a [64-bit] version of the [x86][x] [instruction set
architecture][isa], commonly known as [x86_64].

This Habitat package target is intended for software with older
buildtime or runtime requirements than those supported by the current
`x86_64-linux` package target. Specifically, software with this package
target should run on systems from around the year 2016, when the Habitat
project was publicly launched. This Habitat package target will run on
Linux systems equipped with kernel versions as low as 2.6.32.

[Linux kernel]: https://en.wikipedia.org/wiki/Linux_kernel
[64-bit]: https://en.wikipedia.org/wiki/64-bit_computing
[x]: https://en.wikipedia.org/wiki/X86
[isa]: https://en.wikipedia.org/wiki/Instruction_set_architecture
[x86_64]: https://en.wikipedia.org/wiki/X86-64

References habitat-sh/habitat#5278
Closes habitat-sh/core-plans#1657

### DIFF
--- a/components/core/src/package/target.rs
+++ b/components/core/src/package/target.rs
@@ -280,6 +280,23 @@ supported_package_targets! {
     /// [x86_64]: https://en.wikipedia.org/wiki/X86-64
     ("x86_64-linux", X86_64_Linux, X86_64_LINUX, "x86_64", "linux");
 
+    /// Represents an **older** [Linux kernel]-based system from the 2.6.x family running on a
+    /// [64-bit] version of the [x86][x] [instruction set architecture][isa], commonly known as
+    /// [x86_64].
+    ///
+    /// This Habitat package target is intended for software with older buildtime or runtime
+    /// requirements than those supported by the current `x86_64-linux` package target.
+    /// Specifically, software with this package target should run on systems from around the year
+    /// 2016, when the Habitat project was publicly launched. This Habitat package target will run
+    /// on Linux systems equipped with kernel versions as low as 2.6.32.
+    ///
+    /// [Linux kernel]: https://en.wikipedia.org/wiki/Linux_kernel
+    /// [64-bit]: https://en.wikipedia.org/wiki/64-bit_computing
+    /// [x]: https://en.wikipedia.org/wiki/X86
+    /// [isa]: https://en.wikipedia.org/wiki/Instruction_set_architecture
+    /// [x86_64]: https://en.wikipedia.org/wiki/X86-64
+    ("x86_64-linux-kernel2", X86_64_Linux_Kernel2, X86_64_LINUX_KERNEL2, "x86_64", "linux");
+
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/core/src/package/target.rs
```diff
@@ -280,6 +280,24 @@ supported_package_targets! {
     /// [x86_64]: https://en.wikipedia.org/wiki/X86-64
     ("x86_64-linux", X86_64_Linux, X86_64_LINUX, "x86_64", "linux");
 
+    /// Represents an **older** [Linux kernel]-based system from the kernel version 2.6.x and up
```
scotthain: We should not say "2.6.x and up" - this should state "2.6.x family"

fnichol: I was struggling there for a good way to convey this, will update

fnichol: I also wasn't sure about calling out the magic `2.6.32` number in the second paragraph…problematic or good?

scotthain: Yeah I'd probably remove the magic number, even though we know it works on that version...

scotthain: Actually I take that back - that is actually *the* explicit lowest version we are targeting to support.

fnichol:
Okay, prose updated!

components/core/src/package/target.rs
```diff
@@ -280,6 +280,24 @@ supported_package_targets! {
...
+    /// Represents an **older** [Linux kernel]-based system from the 2.6.x family running on a
+    /// [64-bit] version of the [x86][x] [instruction set architecture][isa], commonly known as
+    /// [x86_64].
+    ///
+    /// This package target represents packages that have certain buildtime or runtime requirements
+    /// that are older than what the more current `x86_64-linux` package target software is
+    /// supporting. Specifically, software with this package target should be able to run on
+    /// systems that were considered current around the 2016 time period (when the Habitat project
+    /// was publicly launched). Linux systems running kernel versions as low as 2.6.32 will run
+    /// these packages.
```
kagarmoe: The first sentence uses `package` 3x, each with a different meaning. Can we find another word? Here are suggestions: 

This package target is intended for packages with specific buildtime or runtime requirements and that are older than what the more current `x86_64-linux` package target software supports. Specifically, software with this package target should be able to run on systems that were current around the 2016, when the Habitat project was publicly launched. These packages will run on Linux systems equipped with kernel versions as low as 2.6.32.

fnichol:
Updated the description based on the great feedback from @kagarmoe.

fnichol:
Another good round of description updates here

PR #44, opened 2018-07-24
Ignore temporary installation directories

stevendanna:
A concurrent change to habitat/habitat-sh will create temporary
directories during package installation with the naming scheme
.hab-pkg-install-RELEASE-RANDOM_ASCII (see habitat-sh/habitat#5369).

Other commands should ignore these directories when generating lists
of releases as they are either:

(1) The result of incomplete package installations that were killed; or,

(2) The result of a concurrently running package installation and will
soon be renamed.

### DIFF
--- a/components/core/src/package/install.rs
+++ b/components/core/src/package/install.rs
@@ -31,6 +31,7 @@ use error::{Error, Result};
 use fs;
 
 pub const DEFAULT_CFG_FILE: &'static str = "default.toml";
+pub const INSTALL_TMP_PREFIX: &'static str = ".hab-pkg-install";
 const PATH_KEY: &'static str = "PATH";
 
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/core/src/package/install.rs
```diff
@@ -691,6 +691,17 @@ impl PackageInstall {
 
         for entry in std::fs::read_dir(version.path())? {
             let entry = entry?;
+            if let Some(path) = entry.path().file_name().and_then(|f| f.to_str()) {
+                if path.starts_with(".hab-pkg-install") {
```
christophermaier: It might be nice to set `".hab-pkg-install"` as a constant here in the core crate. That provides a nice location to clearly document the intention behind it, and also lets us refer to that same constant over in https://github.com/habitat-sh/habitat/pull/5369

stevendanna: 👍 Had meant to do this, thanks for the reminder.

baumanj: Once you address @christophermaier's request, looks good to me

PR #45, opened 2018-07-26
only trim trailing $ from windows account check if it is the local machine account

mwrock:
This was detected in an enterprise environment running the supervisor under a GMSA account (an AD account where AD manages the password). There account names may end in a `$`. Previously I have only seen that with local machine accounts and `LookupAccountNameW` cannot find those accounts with the trailing `$`. However when validating the GMSA account `LookupAccountNameW` will require that trailing `$`. This PR only trims the `$` if it is indeed a local machine account.

We created a small dummy test program onsite to validate this `LookupAccountNameW` behavior. I have tested this locally with made up accounts that end in a `$`.

### DIFF
--- a/components/win-users/src/account.rs
+++ b/components/win-users/src/account.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::env;
 use std::io::Error;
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

PR #46, opened 2018-07-27
Make plan parsing a little more robust.

raskchanky:
Closes https://github.com/habitat-sh/builder/issues/368
Signed-off-by: Josh Black <raskchanky@gmail.com>

### DIFF
--- a/components/core/src/package/plan.rs
+++ b/components/core/src/package/plan.rs
@@ -30,11 +30,27 @@ impl Plan {
         let mut version: Option<String> = None;
         for line in bytes.lines() {
             if let Ok(line) = line {
-                let parts: Vec<&str> = line.splitn(2, "=").collect();
+                // Rather than just blindly accepting values, let's trim all the
+                // whitespace first, verify that we actually have 2 things separated
+                // by an equal sign, and strip out quotes of any kind.
+                //
+                // To do this properly, we probably need some kind of bash parser,
+                // or a plan file syntax that's in a different language that we do
+                // have a parser for (LUA!), but both of those things are beyond the
+                // scope of this task.
+                let parts: Vec<&str> = line.splitn(2, "=").map(|x| x.trim()).collect();
+
+                if parts.len() != 2 {
+                    continue;
+                }
+
+                let mut val = parts[1].replace("\"", "");
+                val = val.replace("'", "");
+
                 match parts[0] {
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/core/src/package/plan.rs
```diff
@@ -30,11 +30,27 @@ impl Plan {
...
+                // have a parser for (LUA!), but both of those things are beyond the
+                // scope of this task.
+                let parts: Vec<&str> = line.splitn(2, "=").map(|x| x.trim()).collect();
+
+                if parts.len() != 2 {
+                    continue;
+                }
+
+                let mut val = parts[1].replace("\"", "");
+                val = val.replace("'", "");
```
stevendanna: This implies that the fields being parsed cannot contain `'` or `"` (i.e. no `pkg_name="jim's-nice-service"`). I think in practice this is fine.

baumanj:
Is there a good reason not to leverage bash itself for this? If you direct the bytes into `bash` but `cat` on a script like this:
```bash
#!/bin/bash

echo "$pkg_name"
echo "$pkg_origin"
echo "$pkg_version"
```
you'd get clean, predictable output in the way that `bash` itself would parse it. For example:
```
➤ cat /Users/jbauman/core-plans/wordpress/plan.sh ../parse-plan.sh | /bin/bash ^/dev/null
wordpress
core
4.7.4
```

christophermaier:
@baumanj I think that opens us up to remote code execution:

`plan.sh`:
```bash
echo "HACKED"
pkg_name="my_package"
# etc...
```

`test.sh`: 
``` bash
#!/bin/bash
echo $pkg_name
```
Execution: 
```bash
cat plan.sh test.sh | bash
HACKED
my_package
```

elliott-davis:
@baumanj @cm we have a similar tool here https://github.com/habitat-sh/core-plans/blob/master/bin/print-deps.sh

baumanj:
Good point, @christophermaier. That raises the question: why are we keeping what is clearly *data*, that needs to be accessed by other parts of the system mixed in with executable code? Have we considered splitting this out?

Another way this bites us is that currently a plan can do runtime things to calculate these values (including sourcing other files). That will work in some parts of our system, but not others. I think it may be smart for long term security and usability to separate immutable data from executable logic in the plan.

PR #51, opened 2018-08-03
[ci] Reimplement lint.sh program for better resilience.

fnichol:
This change reimplements the `support/ci/lint.sh` program which tries to
ensure that any Rust code contributions were prepared with the `rustfmt`
tool. The CI invocation of this programs is identical to the prior
implementation, but there are several advantages to this new approach:

* Drop reliance on flags and options in `rustfmt` to better bridge any
future changes to the program. This program attempts to use `rustfmt`
closer to how a code editor would interact with the formatter.
* Update the exit code checking of `rustfmt`, which has slightly changed
over the last 12+ months.
* Drop any remaining references to `rustfmt` version checks. Our current
approach is to use the version of `rustfmt` which ships as a component
in Rustup alongside a Rust toolchain.
* Expose explicit modes of operation via flags: `all`, `cached`,
`unstaged`, `files`, and `git`.
* Add a `--files` option, allowing easier local lint checking on
specific files.
* Add self-documenting usage via a `--help` flag.

### DIFF
--- a/support/ci/lint.sh
+++ b/support/ci/lint.sh
@@ -12,7 +12,7 @@ main() {
   need_cmd dirname
   need_cmd find
   need_cmd git
-  need_cmd realpath
+  # need_cmd realpath
   need_cmd rustfmt
 
   program="$(basename "$0")"
@@ -249,7 +249,13 @@ lint_file() {
     # Skip files which were deleted
     return 0
   fi
-  if [[ "$(realpath "$_file")" = */target/* ]]; then
+  if echo "$_file" | grep -q '/target/' > /dev/null; then
+    # Would rather do this:
+    #
+    #     if [[ "$(realpath "$_file")" = */target/* ]]; then
+    #
+    # but it doesn't look like realpath is in Travis, and there
+    # appears to be an issue installing it :/
     # Skip files in a `target/` directory
     return 0
   fi
--- a/support/ci/lint.sh
+++ b/support/ci/lint.sh
@@ -20,10 +20,7 @@ main() {
 }
 
 print_help() {
-  need_cmd cat
-
-  cat <<USAGE
-$program
+  echo "$program
 
 Authors: $author
 
@@ -74,7 +71,7 @@ EXAMPLES:
     # two Travis-specific environment variables)
     $program
 
-USAGE
+"
 }
 
 parse_cli_args() {
--- a/support/ci/lint.sh
+++ b/support/ci/lint.sh
@@ -288,7 +288,8 @@ lint_file() {
       ;;
   esac
 
-  if diff --color=always --unified "$_file" "$workdir/$_file" > "$workdir/$_file".diff 2>&1; then
+  # the diff on Travis doesn't seem to support --color=always :|
+  if diff --unified "$_file" "$workdir/$_file" > "$workdir/$_file".diff 2>&1; then
      _diff_exit=0
   else
      _diff_exit="$?"
--- a/support/ci/lint.sh
+++ b/support/ci/lint.sh
@@ -123,7 +123,9 @@ parse_cli_args() {
       -f|--files)
         lint=files
         shift
-        files="$*"
+        # Note: this is effectively a global variable, and is used
+        # later in lint_files
+        files=("$@")
         if [[ -z "$files" ]]; then
           warn "--files option requires one or more file values"
           print_help
@@ -186,28 +188,28 @@ setup() {
 }
 
 lint_files() {
-  local _input_files_cmd _file
+  local _file
 
   case "$lint" in
     all)
-      _input_files_cmd="find . -type f -name '*.rs'"
-      info "Linting all files, selecting files via: '$_input_files_cmd'"
+      readarray -t files < <(find . -type f -name '*.rs')
+      info "Linting all files"
       ;;
     staged)
-      _input_files_cmd="git diff --name-only --cached"
-      info "Linting staged changes, selecting files via: '$_input_files_cmd'"
+      readarray -t files < <(git diff --name-only --cached)
+      info "Linting staged changes"
       ;;
     files)
-      _input_files_cmd="echo '$files'"
-      info "Linting specific files: $files"
+      # files was populated up in parse_cli_args
+      info "Linting files specified with --files"
       ;;
     git)
-      _input_files_cmd="git diff-tree --no-commit-id --name-only -r $git"
-      info "Linting files from Git via: '$_input_files_cmd'"
+      readarray -t files < <(git diff-tree --no-commit-id --name-only -r $git)
+      info "Linting files from Git: $git"
       ;;
     unstaged)
-      _input_files_cmd="git diff --name-only"
-      info "Linting Unstaged changes, selecting files via: '$_input_files_cmd'"
+      readarray -t files < <(git diff --name-only)
+      info "Linting unstaged changes"
       ;;
     *)
       exit_with "Invalid lint type: $lint" 5
@@ -216,7 +218,7 @@ lint_files() {
 
   echo
 
-  eval "$_input_files_cmd" | while read -r _file; do
+  for _file in "${files[@]}"; do
     case "${_file##*.}" in
       rs)
         lint_file "$_file"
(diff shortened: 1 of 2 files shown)


support/ci/lint.sh
```diff
@@ -1,133 +1,361 @@
...
+      lint=git
+      git="${TRAVIS_COMMIT_RANGE:-${TRAVIS_COMMIT:-HEAD}}"
+      info "Selecting files from Git using ref: '$git'"
+    fi
+  else
+    case "$1" in
+      -a|--all)
+        lint=all
+        shift
+        if [[ -n "${1:-}" ]]; then
```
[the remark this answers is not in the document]
baumanj: Another approach would be that instead of checking `$1`, in each individual case, do something like:
```bash
    while [[ -n $1 ]]; do
      if [[ -n "${lint:-}" ]]; then
        warn "Cannot specify both '$lint' mode and $1"
        print_help
        exit_with "Invalid usage" 1
      fi

      case "$1" in
        -a|--all)
          lint=all
          shift
          ;;
        -c|--cached)
          lint=cached
          shift
          ;;
        -f|--files)
          lint=files
          shift
          files="$*"
          while [[ -n $1 ]] && [[ $1 != -* ]]; do
            shift
          done
          ;;
        -g|--git)
          lint=git
          shift
          if [[ -z "${1:-}" ]]; then
            warn "--git option requires a Git ref value"
            print_help
            exit_with "Invalid usage" 1
          fi
          git="$1"
          shift
          ;;
        -h|--help)
          print_help
          exit 0
          ;;
        -u|--unstaged)
          lint=unstaged
          shift
          ;;
      esac
    done
```

support/ci/lint.sh
```diff
@@ -1,133 +1,361 @@
...
+      info "Linting Unstaged changes, selecting files via: '$_input_files_cmd'"
+      ;;
+    *)
+      exit_with "Invalid lint type: $lint" 5
+      ;;
+  esac
+
+  echo
+
+  eval "$_input_files_cmd" | while read -r _file; do
```
baumanj: This construct doesn't work on account of the way read interacts with subshells:

```
$ echo "foo" | read -r bar
$ echo $bar

```
You need to instead do something like:
```
$ read -r bar < <(echo foo)
$ echo $bar
foo
```

Test it out and see, `DEBUG=1 support/ci/lint.sh --files components/core/src/os/mod.r` doesn't end up calling `rustfmt` at all.

I think it would be preferable to avoid all this `eval` complexity entirely in favor of setting arrays. For the regular files case that would just look like:

```bash
files=("$@") # in parse_cli_args
```
for the git cases, something like:
```bash
readarray -t files < <(git diff --name-only) # in lint_files
```
And then 
```bash
for file in "${files[@]}"; do
    lint_file "$file"
done
```

support/ci/lint.sh
```diff
@@ -1,133 +1,361 @@
...
+
+lint_file() {
+  local _file="$1"
+  local _rf_out _rf_exit _diff_out _diff_exit
+
+  if [[ ! -e "$_file" ]]; then
+    # Skip files which were deleted
+    return 0
+  fi
+  if echo "$_file" | grep -q '/target/' > /dev/null; then
```
baumanj: Why not
```bash
  if [[ "$(realpath "$_file")" = */target/* ]]; then 
```
?

support/ci/lint.sh
```diff
@@ -1,133 +1,361 @@
...
+    # Skip files in a `target/` directory
+    return 0
+  fi
+  if [[ "$(basename "$(dirname "$_file")")" == "generated" ]]; then
+    # Skip files directly under a `generated/` directory
+    return 0
+  fi
+
+  need_cmd dirname
```
baumanj: Shouldn't that come before `dirname` is used (on line 259)? And include `basename` as well?

support/ci/lint.sh
```diff
@@ -1,133 +1,361 @@
...
+  need_cmd rustfmt
+
+  info "Running rustfmt on $_file"
+  mkdir -p "$(dirname "$workdir/$_file")"
+
+  set +e
+  _rf_out="$(rustfmt < "$_file" > "$workdir/$_file")"
+  _rf_exit="$?"
+  set -e
```
baumanj: You can avoid the `±e` stuff

Firstly, `_rf_out` won't capture anything. The stdout for `rustfmt` is going to `"$workdir/$_file"` and stderr is not returned from command substitution:
```bash
$ ls bar
ls: cannot access 'bar': No such file or directory

$ ls bar 2>/dev/null

$ _bar_out=$(ls bar)
ls: cannot access 'bar': No such file or directory

$ echo $_bar_out
```

Instead, you could:
```bash
  if rustfmt < "$_file" > "$workdir/$_file" 2> "$workdir/rustfmt_errors"; then
    _rf_exit=0
  else
    _rf_exit="$?"
  fi
```
and remove `_rf_out` in favor of:
```bash
      cat "$workdir/rustfmt_errors"
```

support/ci/lint.sh
```diff
@@ -1,133 +1,361 @@
...
+      # All other exit codes are errors
+      warn "File $_file exited from rustfmt with $_rf_exit"
+      warn "Error output:"
+      echo "$_rf_out"
+      echo "$_file" >> "$failed"
+      return 0
+      ;;
+  esac
+
+  set +e
```
baumanj: Similar comment to above

support/ci/lint.sh
```diff
@@ -1,133 +1,361 @@
...
+      ;;
+  esac
+}
+
+need_cmd() {
+  if ! command -v "$1" > /dev/null 2>&1; then
+    warn "Required command '$1' not found on PATH"
+    exit 127
+  fi
+}
```
baumanj: Instead of always having to remember to put `need_cmd` in the proper places, you could do something like
```bash
try() {
  if command -v "$1" > /dev/null 2>&1; then
    "$@"
  else
    warn "Required command '$1' not found on PATH"
    exit
  fi
}
```
And always write `try rustfmt`.

However, is this really any better than `missing-command: command not found`, which you get for free with `set -e`?

support/ci/lint.sh
```diff
@@ -1,133 +1,361 @@
...
+Authors: $author
+
+Lints Rust source code files based on Git commits.
+
+USAGE:
+    $program [FLAGS] [OPTIONS]
+
+FLAGS:
+    -a, --all         Lints all source files, ignoring status of Git repository
+    -c, --cached      Lints all currently staged files in Git repository
```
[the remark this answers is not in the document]
christophermaier: Looks like this was just echoing the git underlying git option... will take a look at how `--staged` feels in practice.

support/ci/lint.sh
```diff
@@ -1,133 +1,361 @@
...
+}
+
+parse_cli_args() {
+  if [[ -z "${1:-}" ]]; then
+    info "No explicit mode, attempting to auto detect..."
+
+    need_cmd git
+    need_cmd wc
+
+    if [[ $(git diff --name-only | wc -l) -gt 0 ]]; then
```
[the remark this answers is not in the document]
christophermaier: Good catch

support/ci/lint.sh
```diff
@@ -1,133 +1,361 @@
...
+  if [[ -n "${TMPDIR:-}" ]]; then
+    _tmp="${TMPDIR}"
+  elif [[ -d /var/tmp ]]; then
+    _tmp=/var/tmp
+  else
+    _tmp=/tmp
+  fi
+  workdir="$(mktemp -d -p "$_tmp" 2> /dev/null || mktemp -d "${_tmp}/lint.XXXX")"
+  # shellcheck disable=2154
+  trap 'code=$?; rm -rf $workdir; exit $code' INT TERM EXIT
```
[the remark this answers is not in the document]
christophermaier: Hrmm... I could have sworn this wasn't the case in some previous code I've worked on... I'll have to go check that out again, but this is definitely more expected behavior 😄

support/ci/lint.sh
christophermaier: Hrmm... I can run that `lint.sh --files ...` command and it _does_ end up running rustfmt. Operating on arrays does seem cleaner though... will check it out.

support/ci/lint.sh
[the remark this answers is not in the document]
christophermaier: I think a reformulation of these as "subcommands" might be more fruitful, seeing as how they're all mutually exclusive.

I'll leave that for a subsequent PR, though.

PR #52, opened 2018-08-17
Implement is_a_service method for PackageInstall

smacfarlane:
The implementation differs slightly from PackageArchive, due to the differences in how they handle svc_user.

svc_user is widely used through the code base and unifying the behavior for that function is something I think should be done, but would be too complex for this initial use case (habitat-sh/habitat#5298)

I also elected not to remove the `is_runnable` function until habitat-sh/habitat#5298 is merged. 

Signed-off-by: Scott Macfarlane <macfarlane.scott@gmail.com>

### DIFF
--- a/components/core/src/package/install.rs
+++ b/components/core/src/package/install.rs
@@ -194,6 +194,18 @@ impl PackageInstall {
         }
     }
 
+    // hab-plan-build.sh only generates SVC_USER and SVC_GROUP files if it thinks a package is
+    // a service. It determines that by checking for the presence of a run hook file or a
+    // pkg_svc_run value. Therefore, if we can detect the presence of a SVC_USER file, we can
+    // consider this archive a service.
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

baumanj:
Ok, as long as you're aware of the other. Is the long-term plan to eliminate `is_runnable`, or does it have a different use case that we want to keep it around for? Either way, I think a comment about it would be helpful here. Add that, and I'm happy to approve.

Generally, I think relying on an implementation detail of `hab-plan-build.sh` is a bit brittle, but for the near term, it makes sense. For the long term, can we file a separate issue to make the service/binary distinction a well-defined part of the build output? We already have some sort of type indicator, I think.

smacfarlane:
@baumanj    I haven't had the cycles recently to pick this up again.      I'm going to close it out for now and revisit in the future.  Thanks for your assistance!

PR #55, opened 2018-08-28
do not panic validating users and fail validation instead

mwrock:
fixes https://github.com/habitat-sh/habitat/issues/5258

There are some infrequent circumstances where calls to win32 `LookupAccountNameW` will fail for reasons other than the name was not mapped to a user. For example, there may be a brief period of time after a machine has been domain joined where a call might error in:
```
The trust relationship between this workstation and the primary domain failed. (os error 1789)
```
Rather then panicing, we should simply log the error and fail the lookup returning `None`.

### DIFF
--- a/components/win-users/src/account.rs
+++ b/components/win-users/src/account.rs
@@ -80,11 +80,14 @@ fn lookup_account(name: &str, system_name: Option<String>) -> Option<Account> {
     match Error::last_os_error().raw_os_error().unwrap() as u32 {
         ERROR_INSUFFICIENT_BUFFER => {}
         ERROR_NONE_MAPPED => return None,
-        _ => panic!(
-            "Error while looking up account for {}: {}",
-            name,
-            Error::last_os_error()
-        ),
+        _ => {
+            info!(
+                "Error while looking up account for {}: {}",
+                name,
+                Error::last_os_error()
+            );
+            return None;
+        }
     }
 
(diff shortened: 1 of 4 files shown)


thesentinels:

Thank you for contributing!

components/win-users/src/account.rs
```diff
@@ -103,11 +106,12 @@ fn lookup_account(name: &str, system_name: Option<String>) -> Option<Account> {
         )
     };
     if ret == 0 {
-        panic!(
+        info!(
```
baumanj: ditto

components/win-users/src/account.rs
```diff
@@ -80,11 +80,14 @@ fn lookup_account(name: &str, system_name: Option<String>) -> Option<Account> {
     match Error::last_os_error().raw_os_error().unwrap() as u32 {
         ERROR_INSUFFICIENT_BUFFER => {}
         ERROR_NONE_MAPPED => return None,
-        _ => panic!(
-            "Error while looking up account for {}: {}",
-            name,
-            Error::last_os_error()
-        ),
+        _ => {
+            info!(
```
baumanj: If this was previously a panic, is it at least worth `error!` rather than `info!`?

mwrock: YES! I don't think I knew about that macro or forgot it. will change

PR #70, opened 2018-10-06
rework permissions into separate APIs for linux and windows

### LINKED ISSUE #57: Remove misleading implementations of set_permissions and set_owner on Windows
Currently [`set_permissions` and `set_owner`](https://github.com/habitat-sh/core/blob/master/components/core/src/util/perm.rs#L22-L100) depend upon [`chmod`](https://github.com/habitat-sh/core/blob/master/components/core/src/util/perm.rs#L90) and [`chown`](https://github.com/habitat-sh/core/blob/master/components/core/src/util/perm.rs#L65). However, [on Windows, these functions are implemented as essentially no-ops](https://github.com/habitat-sh/core/blob/master/components/core/src/os/filesystem/windows.rs#L28-L34). This leads to code in `habitat` and `builder` which is potentially very misleading.

Once the dependent code is fixed (see https://github.com/habitat-sh/habitat/pull/5591 and https://github.com/habitat-sh/builder/pull/627), the `set_permissions` and `set_owner` implementations should be removed from the Windows platform until we can provide proper implementations. This will require additional, nontrivial FFI work since Window's permissions and ownership models are different.

mwrock:
This fixes #57 

This provides a separate API altogether for Windows. Because the underlying mechanisms of manipulating permissions are so different across the two platforms, having a single API is not feasible.


thesentinels:

Thank you for contributing!

Issue #11, opened 2018-03-02
Signals in rapid succession could end up unhandled


baumanj:
The [`handle_signal`](https://github.com/habitat-sh/core/blob/master/components/core/src/os/signals/unix.rs#L38) function is registered to receive signals, but all it does currently is modify two static atomic variables to indicate that a signal is waiting to be processed and which signal it is. Consumer code calls [`check_for_signal`](https://github.com/habitat-sh/core/blob/master/components/core/src/os/signals/unix.rs#L51), which resets the state of these variables before returning the type of signal that is waiting (or `None`).

In the event that multiple signals are received before `check_for_signal` is called, all but the last would be functionally ignored. Additionally, there is no way for the code calling `check_for_signal` to indicate whether it's successfully processed the signal.

baumanj:
Some useful info:
http://www.microhowto.info/howto/reap_zombie_processes_using_a_sigchld_handler.html
http://pubs.opengroup.org/onlinepubs/9699919799/functions/wait.html#tag_16_645
http://pubs.opengroup.org/onlinepubs/9699919799/functions/signal.html
http://pubs.opengroup.org/onlinepubs/9699919799/functions/sigaction.html

baumanj:
Confirmed valid in v0.57

PR #75, opened 2018-10-11
When multiple signals are received, return them in order; don't drop any

baumanj:
Resolves https://github.com/habitat-sh/core/issues/11

With the supervisor running, if we execute
```
➤ pkill -USR1 hab-sup; pkill -USR2 hab-sup
```
Before:
```
Received invalid signal: #12
```
After:
```
Received invalid signal: #10
Received invalid signal: #12
```

### DIFF
--- a/components/core/src/os/signals/unix.rs
+++ b/components/core/src/os/signals/unix.rs
@@ -14,18 +14,18 @@
 
 //! Traps and notifies UNIX signals.
 
-use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering, ATOMIC_BOOL_INIT, ATOMIC_USIZE_INIT};
-use std::sync::{Once, ONCE_INIT};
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

baumanj:
In general a `Mutex` is simpler and so I would prefer it unless there's a compelling reason for the added complexity of a `RwLock`. Generally that would be in the form of high contention for the lock and a large ratio of readers:writers. In this case there should be essentially zero contention as the only code paths that access the mutex are the `check_for_signal` callers, which typically run ~1x/second, always from the same thread, and the signal interrupt, which only runs when there are new signals. Also, there are no read-only consumers: `handle_signal` always writes, and `check_for_signal` would need a write lock any time the Vector was non-empty. Since there's no read → write lock upgrade, the code would be unnecessarily complex anyway.

PR #81, opened 2018-11-01
Add fs::AtomicWriter and fs::atomic_write

stevendanna:
The Habitat code has a number of places where it writes to a temporary
file and then renames the file. The goal of this pattern is to achieve
an atomic write to the file to prevent concurrent readers from seeing
incomplete versions of this file.

However, getting this pattern right can be tricky. For example, the
temporary file must be created in the same directory to ensure an
atomic rename is possible and the temporary file must by fsync'd to
disk before the rename to protect against seeing partial files in some
powerloss scenarios.

The AtomicWriter is an attempt to help us get this pattern right
across the code base.

### DIFF
--- a/components/core/src/fs.rs
+++ b/components/core/src/fs.rs
@@ -18,6 +18,7 @@ use std::fs as stdfs;
 use std::io::{self, Write};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
+use tempfile;
 
 use crate::env as henv;
 use crate::error::{Error, Result};
@@ -654,6 +655,85 @@ fn fs_root_path() -> PathBuf {
     }
 }
 
+/// parent returns the parent directory of the given path, accounting
+/// for the fact that a relative path with no directory separator
+/// returns an empty parent. This function transforms it to return "."
+/// so that it can more easily be used in functions that expect to
+/// take a directory.
+fn parent(p: &Path) -> io::Result<&Path> {
+    match p.parent() {
+        Some(parent_path) if parent_path.as_os_str().is_empty() => Ok(&Path::new(".")),
+        Some(nonempty_parent_path) => Ok(nonempty_parent_path),
+        None => Err(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            "path has no parent",
+        )),
+    }
+}
+
+/// An AtomicWriter atomically writes content to a file at the
+/// specified path using a tempfile+rename strategy to achieve
+/// atomicity.
+///
+/// The goal of this function is to ensure that other observers can
+/// only ever see:
+///
+/// - Any previous file at that location
+/// - The new file with your content
+///
+/// without seeing any possible intermediate state.
+///
+/// Assumes that the parent directory of dest_path exists.
+///
+pub struct AtomicWriter {
+    dest: PathBuf,
+    tempfile: tempfile::NamedTempFile,
+}
+
+impl AtomicWriter {
+    pub fn new(dest_path: &Path) -> io::Result<Self> {
+        let parent = parent(dest_path)?;
+        let tempfile = tempfile::NamedTempFile::new_in(parent)?;
+        Ok(Self {
+            dest: dest_path.to_path_buf(),
+            tempfile: tempfile,
+        })
+    }
+
+    pub fn with_writer<F, T, E>(mut self, op: F) -> std::result::Result<T, E>
+    where
+        F: FnOnce(&mut std::fs::File) -> std::result::Result<T, E>,
+        E: From<std::io::Error>,
+    {
+        let r = op(&mut self.tempfile.as_file_mut())?;
+        self.finish()?;
+        Ok(r)
+    }
+
+    /// finish completes the atomic write by calling sync on the
+    /// temporary file to ensure all data is flushed to disk and then
+    /// renaming the file into place.
+    fn finish(self) -> io::Result<()> {
+        self.tempfile.as_file().sync_all()?;
+        debug!(
+            "Renaming {} to {}",
+            self.tempfile.path().to_string_lossy(),
+            &self.dest.to_string_lossy()
+        );
+        stdfs::rename(self.tempfile.path(), &self.dest)
+        // NOTE(ssd) 2018-11-01: On some filesystems we should fsync
+        // the parent directory here if we wanted this file to also
+        // provide a durability guarantee.
+    }
+}
+
+/// atomic_write is a helper function for the most common use of
+/// AtomicWriter.
+pub fn atomic_write(dest_path: &Path, data: impl AsRef<[u8]>) -> io::Result<()> {
+    let w = AtomicWriter::new(dest_path)?;
+    w.with_writer(|f| f.write_all(data.as_ref()))
+}
+
 #[cfg(test)]
 mod test_find_command {
 
@@ -834,5 +914,53 @@ mod test_find_command {
                 assert_eq!(found_path, target_path);
             }
         }
+    }
+}
+
+#[cfg(test)]
+mod test_atomic_writer {
+    use super::{atomic_write, AtomicWriter};
+    use std::fs::{remove_file, File};
+    use std::io::{Read, Seek, SeekFrom, Write};
+    use std::panic;
+    use tempfile;
+
+    const EXPECTED_CONTENT: &str = "A very good file format";
+
+    #[test]
+    fn atomic_write_writes_file() {
+        let dest_file = tempfile::NamedTempFile::new().expect("could not create temp file");
+        let dest_file_path = dest_file.path();
+        remove_file(dest_file_path).expect("could not delete temp file");
+
+        let res = atomic_write(dest_file_path, EXPECTED_CONTENT);
+        assert!(res.is_ok());
+        let mut f = File::open(dest_file_path).expect("file not found");
+        let mut actual_content = String::new();
+        f.read_to_string(&mut actual_content)
+            .expect("failed to read file");
+        assert_eq!(EXPECTED_CONTENT, actual_content);
+    }
+
+    #[test]
+    fn with_atomic_writer_writes_file() {
+        let dest_file = tempfile::NamedTempFile::new().expect("could not create temp file");
+        let dest_file_path = dest_file.path();
+        remove_file(dest_file_path).expect("could not delete temp file");
+
+        let w = AtomicWriter::new(dest_file_path).expect("could not create AtomicWriter");
+        let res = w.with_writer(|writer| {
+            writer.write_all(b"not the right content")?;
+            writer.seek(SeekFrom::Start(0))?;
+            writer.write_all(EXPECTED_CONTENT.as_bytes())?;
+            writer.flush() // not needed, just making sure we can call it
+        });
+        assert!(res.is_ok());
+
+        let mut f = File::open(dest_file_path).expect("file not found");
+        let mut actual_content = String::new();
+        f.read_to_string(&mut actual_content)
+            .expect("failed to read file");
+        assert_eq!(EXPECTED_CONTENT, actual_content);
     }
 }


thesentinels:

Thank you for contributing!

components/core/src/fs.rs
```diff
@@ -433,6 +435,112 @@ fn fs_root_path() -> PathBuf {
...
+///
+/// Assumes that the parent directory of dest_path exists.
+///
+pub struct AtomicWriter {
+    dest: PathBuf,
+    parent: PathBuf,
+    tmp_file: tempfile::NamedTempFile,
+}
+
+impl AtomicWriter {
```
stevendanna: Reviewer Note: I had originally wanted an API in which the caller just passed a closure into a `write` callback.  I found this became pretty verbose at the call site because of all the type annotations required to get it right, but perhaps I was just doing something incorrectly.

However, even after changing to this API, I found it was still rather annoying to use if I didn't return std::io::Result's in most cases.

I'm sure this is solvable if y'all would prefer something different.

components/core/src/fs.rs
```diff
@@ -433,6 +435,112 @@ fn fs_root_path() -> PathBuf {
...
+            tmpfile.path().to_string_lossy()
+        );
+        return Ok(AtomicWriter {
+            dest: dest_path.to_path_buf(),
+            parent: parent.to_path_buf(),
+            tmp_file: tmpfile,
+        });
+    }
+
+    pub fn finish(&self) -> ::std::io::Result<()> {
```
stevendanna: I liked this as an explicit function rather than tying it to Drop since the caller probably wants to be able to handle the error.  If we are worried about people using the writer after calling finish, perhaps we can track a bit of state and make every other function return an error if that happens.

components/core/src/fs.rs
```diff
@@ -433,6 +435,112 @@ fn fs_root_path() -> PathBuf {
...
+        );
+        fs::rename(self.tmp_file.path(), &self.dest)?;
+
+        // On unix, also sync the directory. This helps make sure our
+        // rename is durable in the case of powerloss after this
+        // function returns. However, since this is not supported on
+        // all filesystems (for example: virtualbox fs), we ignore the
+        // error.
+        #[cfg(unix)]
+        match fs::File::open(&self.parent).and_then(|f| f.sync_all()) {
```
stevendanna: We can remove this if y'all would prefer. This impacts durability but not atomicity as far as I understand the situation.

components/core/src/fs.rs
```diff
@@ -433,6 +435,112 @@ fn fs_root_path() -> PathBuf {
...
+    tmp_file: tempfile::NamedTempFile,
+}
+
+impl AtomicWriter {
+    pub fn new(dest_path: &Path) -> ::std::io::Result<AtomicWriter> {
+        let parent = parent(dest_path).ok_or(::std::io::Error::new(
+            ::std::io::ErrorKind::InvalidInput,
+            "path has no parent",
+        ))?;
+        let tmpfile = tempfile::NamedTempFile::new_in(parent)?;
```
stevendanna: On linux it should be possible to use O_TEMPFILE to get a file not on disk and then use linkat if we want to get fancy, but this is closer to what all the callsites where doing before.

components/core/Cargo.toml
```diff
@@ -28,6 +28,7 @@ serde_derive = "*"
 serde_json = "*"
 sodiumoxide = "0.0.16"
 tempdir = "*"
+tempfile = "*"
```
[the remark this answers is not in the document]
christophermaier: Looks like `tempdir` is being subsumed into `tempfile` anyway: https://github.com/rust-lang-deprecated/tempdir

In light of that, using `tempfile` is 👍 ; I'll file an issue to go through and migrate existing uses of `tempdir` to `tempfile`.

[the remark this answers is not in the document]
christophermaier: Here's the master issue for the `tempdir`/`tempfile` cleanup.

https://github.com/habitat-sh/habitat/issues/5806

components/core/src/fs.rs
```diff
@@ -433,6 +435,112 @@ fn fs_root_path() -> PathBuf {
...
+        match fs::File::open(&self.parent).and_then(|f| f.sync_all()) {
+            Ok(_) => {}
+            Err(err) => debug!("Failed to sync directory: {:?}", err),
+        }
+
+        return Ok(());
+    }
+}
+
+impl Write for AtomicWriter {
```
baumanj: A somewhat more flexible approach here would be to allow `AtomicWriter` to be dereferenced to the backing file:

```rust
impl ::std::ops::Deref for AtomicWriter {
    type Target = fs::File;

    fn deref(&self) -> &Self::Target {
        self.tempfile.as_file()
    }
}

impl ::std::ops::DerefMut for AtomicWriter {
    fn deref_mut(&mut self) -> &mut fs::File {
        self.tempfile.as_file_mut()
    }
}
```

That obviates the need to implement any additional traits for `AtomicWriter` and it's a pattern we use elsewhere.

The main argument against is that it seems like [potentially bad style in rust](http://doc.rust-lang.org/1.30.0/std/ops/trait.Deref.html):
> Implementing Deref for smart pointers makes accessing the data behind them convenient, which is why they implement Deref. On the other hand, the rules regarding Deref and DerefMut were designed specifically to accommodate smart pointers. Because of this, Deref should only be implemented for smart pointers to avoid confusion.

Then again, one could make the argument that `AtomicWriter` _is_ a start pointer of a sort.

components/core/src/fs.rs
```diff
@@ -433,6 +435,112 @@ fn fs_root_path() -> PathBuf {
...
+
+impl AtomicWriter {
+    pub fn new(dest_path: &Path) -> ::std::io::Result<AtomicWriter> {
+        let parent = parent(dest_path).ok_or(::std::io::Error::new(
+            ::std::io::ErrorKind::InvalidInput,
+            "path has no parent",
+        ))?;
+        let tmpfile = tempfile::NamedTempFile::new_in(parent)?;
+        debug!(
+            "New atomic writer created for {}. Parent is {}, Underying tempfile is {}",
```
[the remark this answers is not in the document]
stevendanna: Might just remove this log line unless you think it is helpful.  It is mostly cruft from local development.

components/core/src/fs.rs
baumanj: Upon further consideration, probably not worth it since it make the calls like
```rust
BroadcastWriter::new(&mut w, progress)
```
need to be
```rust
BroadcastWriter::new(&mut *w, progress)
```

Unless we find ourselves needing to add a bunch more traits here, I think the original approach is cleaner.

components/core/src/fs.rs
stevendanna: > I'm not sure what most callers would usefully do with the error other than log it, so I'm more in favor of having this happen on drop.

If this function fails, it likely means the file you wanted to exist doesn't necessarily exist, so I would hope that at least a few callers are propogating that back up to users.

baumanj: I would be an favor of a `drop` implementation that panicked if `finish` were not called before drop since that's clearly a logic error.

components/core/src/fs.rs
stevendanna: @christophermaier Well, I learned something interesting trying to implement your suggestion:

```
error: attributes are not yet allowed on `if` expressions                                                                                             
```

😂

components/core/src/fs.rs
```diff
@@ -433,6 +435,112 @@ fn fs_root_path() -> PathBuf {
...
+// parent return the parent directory, accounting for the fact that a
+// relative path with no directory separator returns an empty path.
+// This function transforms it to return "." so that it can more
+// easily be used in functions that expect to take a directory.
+fn parent(p: &Path) -> Option<&Path> {
+    match p.parent() {
+        None => None,
+        Some(parent_path) if parent_path.as_os_str().is_empty() => Some(&Path::new(".")),
+        parent_path => parent_path,
+    }
```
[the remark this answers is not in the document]
stevendanna: Erlang has permanently corrupted my brain so I went with your first suggestion :D.

components/core/src/fs.rs
stevendanna: I've elided this for now and left a comment

stevendanna:
Thanks for the review! I took care of the minor changes.  I'll take a look at the situation wish finish/Drop in a bit.

components/core/src/fs.rs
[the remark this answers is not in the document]
baumanj: Happy either way

components/core/src/fs.rs
```diff
@@ -433,6 +435,95 @@ fn fs_root_path() -> PathBuf {
...
+        // NOTE(ssd) 2018-11-01: On some filesystems we should fsync
+        // the parent directory here if we wanted this file to also
+        // provide a durability guarantee.
+        return Ok(());
+    }
+}
+
+impl Write for AtomicWriter {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        self.tempfile.as_file().write(buf)
```
baumanj: On account of https://github.com/habitat-sh/habitat/pull/5804, @christophermaier  and I have been talking about logging all writes at an `info!` level. Maybe something like:
```rust
        info!(
            "Writing {} bytes to {}",
            buf.len(),
            self.dest.to_string_lossy()
        );
```

components/core/src/fs.rs
```diff
@@ -433,6 +435,95 @@ fn fs_root_path() -> PathBuf {
...
+
+impl Seek for AtomicWriter {
+    fn seek(&mut self, pos: SeekFrom) -> io::Result<u64> {
+        self.tempfile.as_file().seek(pos)
+    }
+}
+
+/// atomic_write_bytes is a helper function for the most common use of
+/// AtomicWriter
+pub fn atomic_write_bytes(dest_path: &Path, bytes: &[u8]) -> io::Result<()> {
```
baumanj: Since the callers to this function (as of https://github.com/habitat-sh/habitat/pull/5803) have `String`s rather than `&[u8]`s, how about making this function take a `String` (or `&str`) and do the bytes conversion here. Or, if we want to get a little fancy, we could make it generic:

```rust
    pub fn atomic_write<'a, T>(dest_path: &Path, data: T) -> io::Result<()>
    where
        T: ::std::convert::Into<&'a [u8]>,
    {
        let mut w = Self::new(dest_path)?;
        w.write_all(data.into())?;
        w.finish()?;
        Ok(())
    }
```

I think this implementation may need some tweaking, but it's late in the day for me and I think you get the idea.

stevendanna:
Apologies that this has sat around for so long.  We've had another user hit a bug related to missing fsync's which reminded me that this was still outstanding.

I force-pushed to get rid of the conflicts so it might be hard to see what has changed since this was last reviewed:

1) Merge conflicts resulting from changed imports have been resolved
2) Clippy violations have been fixed up
3) I've implemented the "operations fail after finish is called" by wrapping the tempfile in Option as recommended by @christophermaier 
4) I've implemented Drop for AtomicWriter that panic's if AtomicWriter is called without finish being called as recommended by @baumanj 
5) I've added tests for (3) and (4)
6) I've changed `atomic_write_bytes` to `atomic_write` and it now takes a `&str`.  I tried implementing the more generic version using `::std::convert::Into` but found that neither std::string nor &str actually implemented the required traits to take advantage of it.  Perhaps I was doing something incorrectly, I don't program in Rust every day, so user error is the most likely blocker here :D.

This will require changing the calling code; however, I'm currently finding it difficult to build hab on Fedora.

In cleaning this up, I found two things that might be good follow ups:

- The tempfile library seems to have a persist() method that we can use to avoid named temporary files in some case.  However, that persist() function doesn't call fsync so we'd still need most of this code.

- I've experimented with adding the directory sync back into this code and simply ignoring the EINVAL we get back in the case where it isn't implemented.  However, since the current code in habitat-sh/habitat doesn't do the fsync's either, I'm happy to keep that out of this PR since this is still moving us in the right direction.

components/core/src/fs.rs
```diff
@@ -654,6 +655,132 @@ fn fs_root_path() -> PathBuf {
...
+///
+/// Assumes that the parent directory of dest_path exists.
+///
+pub struct AtomicWriter {
+    dest: PathBuf,
+    tempfile: Option<tempfile::NamedTempFile>,
+}
+
+impl AtomicWriter {
+    pub fn new(dest_path: &Path) -> io::Result<AtomicWriter> {
```
christophermaier: Minor, but this would probably be more idiomatically called something like `from_path`, since `new` is generally for no-argument functions that return `Self`.

Not a big deal, though.

components/core/src/fs.rs
```diff
@@ -654,6 +655,132 @@ fn fs_root_path() -> PathBuf {
...
+
+                // NOTE(ssd) 2018-11-01: On some filesystems we should fsync
+                // the parent directory here if we wanted this file to also
+                // provide a durability guarantee.
+                self.tempfile = None;
+                return Ok(());
+            }
+            None => return Ok(()),
+        }
+    }
```
christophermaier: What you have here is fine, but can be more concisely and equivalently formulated as:

```rust
    pub fn finish(&mut self) -> io::Result<()> {
        if let Some(t) = self.tempfile.take() {
            t.as_file().sync_all()?;
            debug!(
                "Renaming {} to {}",
                t.path().to_string_lossy(),
                &self.dest.to_string_lossy()
            );
            stdfs::rename(t.path(), &self.dest)?;
            // NOTE(ssd) 2018-11-01: On some filesystems we should fsync
            // the parent directory here if we wanted this file to also
            // provide a durability guarantee.
        }
        Ok(())
    }
```
The key enabler of this is [std::option::Option#take](http://doc.rust-lang.org/1.32.0/std/option/enum.Option.html#method.take), which gives you the original value and leaves a `None` in its place.

components/core/src/fs.rs
```diff
@@ -654,6 +655,132 @@ fn fs_root_path() -> PathBuf {
...
+    fn drop(&mut self) {
+        if self.tempfile.is_some() {
+            panic!("AtomicWriter dropped before finish() was called")
+        }
+    }
+}
+
+/// atomic_write is a helper function for the most common use of
+/// AtomicWriter
+pub fn atomic_write(dest_path: &Path, data: &str) -> io::Result<()> {
```
christophermaier: Are clients ever intended to explicitly create an `AtomicWriter` on their own, or will they just be using this function?

If the latter, then we could only expose this function as public, and make `AtomicWriter` a private implementation detail.

components/core/src/fs.rs
stevendanna: Ah!  Thanks, I had a note to myself to look into what the `new` vs `from` distinction was in the various stdlib functions.

components/core/src/fs.rs
stevendanna: Awesome, take looks very useful

components/core/src/fs.rs
stevendanna: Some people need to make their own AtomicWriter because they want to write their custom file formats in chunks and then seek into the file.  I thought about changing that to write into an in-memory buffer and then just using this interface, but...at this point I can't remember why I didn't do that.  I might have just gotten nervous about the amount of change I was introducing at once.

christophermaier: We can evolve things as we go. What's here is certainly valuable enough as it is!

components/core/src/fs.rs
baumanj: I think `new` with arguments is fine (the stdlib does it plenty). To me, what `new` implies is that it's the canonical way to create an instance. If `new` has arguments, it would be strange for there to be a nullary function that also created an instance. Also, `new` should return `Self`, but I think `Result<Self>` is ok too. See [this discussion](https://github.com/habitat-sh/habitat/pull/6100/files#r252381685) if you're curious.

components/core/src/fs.rs
```diff
@@ -654,6 +655,132 @@ fn fs_root_path() -> PathBuf {
...
+/// parent returns the parent directory of the given path, accounting
+/// for the fact that a relative path with no directory separator
+/// returns an empty parent. This function transforms it to return "."
+/// so that it can more easily be used in functions that expect to
+/// take a directory.
+fn parent(p: &Path) -> Option<&Path> {
+    match p.parent() {
+        Some(parent_path) if parent_path.as_os_str().is_empty() => Some(&Path::new(".")),
+        nonempty_parent_path_or_none => nonempty_parent_path_or_none,
+    }
```
baumanj: I realize nothing has changed here, but it occurs to me that as long as we're making a separate function for this, we could do the error handling here:
```rust
fn parent(p: &Path) -> io::Result<&Path> {
    match p.parent() {
        Some(parent_path) if parent_path.as_os_str().is_empty() => Ok(&Path::new(".")),
        Some(nonempty_parent_path) => Ok(nonempty_parent_path),
        None => Err(io::Error::new(
            io::ErrorKind::InvalidInput,
            "path has no parent",
        )),
    }
```
That would nicely simplify the code that's consuming it from:
```rust
        let parent = parent(dest_path)
            .ok_or_else(|| io::Error::new(io::ErrorKind::InvalidInput, "path has no parent"))?;
```
to
```rust
        let parent = parent(dest_path)?;
```

components/core/src/fs.rs
```diff
@@ -654,6 +655,132 @@ fn fs_root_path() -> PathBuf {
...
+            Some(t) => t.as_file().flush(),
+            None => Err(io::Error::new(
+                io::ErrorKind::Other,
+                "flush called on AtomicWriter after finish",
+            )),
+        }
+    }
+}
+
+impl Seek for AtomicWriter {
```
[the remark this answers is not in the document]
stevendanna: We do, here: https://github.com/habitat-sh/habitat/blob/master/components/butterfly/src/rumor/dat_file.rs#L262-L264

I think we could remove it by writing the dat_file to an in-memory buffer first, but I wanted to limit the scope of the change for now.

components/core/src/fs.rs
stevendanna: 👍 will do

components/core/src/fs.rs
stevendanna: Nice

components/core/src/fs.rs
stevendanna: `fsync()` semantics vary so much depending on filesystem and OS version in terms of what subsequent calls will do, I think not retrying is the right thing to do.

> Really though, I don't see a situation where there's a lot a consumer could reasonably do with an error return from `AtomicWriter::finish` other than log and propagate the failure.

Yup, the kind of errors that are likely to result in errors in finish() are going to likely require user-intervention to resolve (full disk, disk i/o errors, etc).
I

components/core/src/fs.rs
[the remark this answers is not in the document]
baumanj: Oh, right. Well, that's fine I just wanted to check.

stevendanna:
I've added the drop implementation, but I think it is a bit problematic.  Testing it now, but I'm pretty sure it will pean that the user gets a panic anytime a write error happens and the caller propogates that up the call stack.

stevendanna:
Yeah, let's hold on for a minute on this.  I'll look at this a bit more tomorrow.  Thanks for the review.

baumanj:
> I've added the drop implementation, but I think it is a bit problematic. Testing it now, but I'm pretty sure it will pean that the user gets a panic anytime a write error happens and the caller propogates that up the call stack.

I think that shouldn't be the case if you use the `Option::take` approach; `self.tempfile` should get set to `None` before any of the possible errors.

stevendanna:
> I think that shouldn't be the case if you use the `Option::take` approach; `self.tempfile` should get set to `None` before any of the possible errors.

True, but that is only for finish. write, sync, and flush can all fail as well.

stevendanna:
And in those cases I'm less sure you want to lose track of that underlying tempfile 🤔

baumanj:
> And in those cases I'm less sure you want to lose track of that underlying tempfile 🤔

🤦‍♂️ right, good point.

I think perhaps it could be restructured such that any errors encountered are stored internally (and further IO ops become no-ops) until `finish` is called.

stevendanna:
As an update on this.  With some help from @baumanj I'm working on getting the original closure-based API working so we can avoid these shenanigans.

stevendanna:
There is probably some renaming to happen, but here is the closure-based API.  Users no longer call `finish` themselves and thus we don't have to worry about all the issues around drop.

components/core/src/fs.rs
```diff
@@ -654,6 +655,118 @@ fn fs_root_path() -> PathBuf {
...
+        );
+        stdfs::rename(self.tempfile.path(), &self.dest)
+        // NOTE(ssd) 2018-11-01: On some filesystems we should fsync
+        // the parent directory here if we wanted this file to also
+        // provide a durability guarantee.
+    }
+}
+
+/// HiddenWriter impements Seek+Write without giving access to the
+/// underlying std::fs::File
```
baumanj: Why are we trying to restrict access to the underlying file? And why add `HiddenWriter` when `AtomicWriter` is basically the same thing?

stevendanna: > Why are we trying to restrict access to the underlying file?

My thinking here was that we might not want callers depending on the fact that it is a std::fs::File specifically.  For instance, we likely wouldn't want them transforming it into a raw FD and doing something with it that could prevent finish() from working.

> And why add `HiddenWriter` when `AtomicWriter` is basically the same thing?

I suppose I could just leave these methods on AtomicWriter and pass that in, but then I think you'd be able to call with_writer recursively which could be wierd.  Perhaps the compiler would disallow it, I didn't check.  Basically, what I wanted here, is for callers to *just* be able to depend on Write+Seek

baumanj: I feel we don't need to protect against that level of shenanigans. If the consumers do weird stuff, that's on them. Previously I was trying to protect against people forgetting to call `finish`, which is an innocent mistake, I totally expect people to make. The closure approach handles that beautifully, but I don't think we need to treat the consumers like adversaries. Providing them with a `&mut File` should be simple and reasonable, I think.

stevendanna: Okie dokie.  Took it out.

baumanj:
This looks great! Shall we move on to https://github.com/habitat-sh/habitat/pull/5803?

PR #84, opened 2018-11-01
Add a natural ordering of PackageIdents

jamesc:
It matches by the parts of PackageIdent, using our custom
`version` matches and lexicographic ordering otherwise

NOTE.  Out of scope is looking at the current implementation of Ord/PartialOrd/PartialEq which is confusing and gives some concerns (I don't think the generated PartialEq will match PartialOrd) but I'd rather tackle replacing that separately as there are insufficient tests around it to just replace.

Signed-off-by: James Casey <james@chef.io>

### DIFF
--- a/components/core/src/package/ident.rs
+++ b/components/core/src/package/ident.rs
@@ -137,6 +137,31 @@ impl PackageIdent {
         }
     }
 
+    /// Compare two `PackageIdent`s component by component:
+    /// i.e. start with origin, then name, then version, then
+    /// release. The first component to be not equal, then return
+    /// the greater/lesser
+    ///
+    /// TODO: This should probably be the natural implementation of `Ord::cmp`.
+    /// To be investigated why we have a different implementation
+    pub fn by_parts_cmp(self: &PackageIdent, other: &PackageIdent) -> Ordering {
+        if self.origin != other.origin {
+            return self.origin.cmp(&other.origin);
+        }
+
+        if self.name != other.name {
+            return self.name.cmp(&other.name);
+        }
+        match version_sort(
+            self.version.as_ref().unwrap(),
+            other.version.as_ref().unwrap(),
+        ) {
+            Ok(Ordering::Equal) => self.release.cmp(&other.release),
+            Ok(ordering) => ordering,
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/core/src/package/ident.rs
```diff
@@ -137,6 +137,31 @@ impl PackageIdent {
...
+        if self.name != other.name {
+            return self.name.cmp(&other.name);
+        }
+        match version_sort(
+            self.version.as_ref().unwrap(),
+            other.version.as_ref().unwrap(),
+        ) {
+            ord @ Ok(Ordering::Greater) | ord @ Ok(Ordering::Less) => ord.unwrap(),
+            Ok(Ordering::Equal) => self.release.cmp(&other.release),
+            Err(_) => Ordering::Less,
```
christophermaier: The match arms could be simplified to

``` rust
match ... {
  Ok(Ordering::Equal) => self.release.cmp(&other.release),
  Ok(ordering) => ordering,
  Err(_) => Ordering::Less,
}
```

components/core/src/package/ident.rs
```diff
@@ -137,6 +137,31 @@ impl PackageIdent {
         }
     }
 
+    /// Compare two `PackageIdent`s component by component:
+    /// i.e. start with origin, then name, then version, then
+    /// release. The first component to be not equal, then return
+    /// the greater/lesser
+    ///
+    /// TODO: This should probably be the natural implementation of `Ord::cmp`.
+    /// To be investigated why we have a different implementation
```
christophermaier: I don't recall the exact reason for this... I wonder if it's somehow related to how we refer to services that have been loaded by their name, rather than their origin and name.

jamesc: Yeah. that's what I guessed might be the usage.  If so, following @baumanj suggestion it might be better to use a custom implementation and make the default this natural ordering

christophermaier: Yeah, if that's the case, then making this the default and special-casing the other seems like the way to go.

PR #86, opened 2018-11-06
 prepend fs_root to environment_for_command

mwrock:
All upstream usages of `environment_for_command` end up prepending the `FS_ROOT`. I was about to add a [new usage](https://github.com/habitat-sh/habitat/pull/5813) and so it seems best to add it here.

### DIFF
--- a/components/core/src/fs.rs
+++ b/components/core/src/fs.rs
@@ -197,6 +197,22 @@ where
     pkg_path
 }
 
+/// Given a linux style absolute path (prepended with '/') and a fs_root,
+/// this will "re-root" the path just under the fs_root. Otherwise returns
+/// the given path unchanged. Non-Windows platforms will always return the
+/// unchanged path.
+pub fn fs_rooted_path<T>(path: &PathBuf, fs_root: Option<T>) -> PathBuf
+where
+    T: AsRef<Path>,
+{
+    match fs_root {
+        Some(ref root) if path.starts_with("/") && cfg!(windows) => {
+            Path::new(root.as_ref()).join(path.strip_prefix("/").unwrap())
+        }
+        _ => path.to_path_buf(),
+    }
+}
+
 /// Returns the absolute path for a given command, if it exists, by searching the `PATH`
 /// environment variable.
 ///
--- a/components/core/src/package/install.rs
+++ b/components/core/src/package/install.rs
@@ -237,12 +237,22 @@ impl PackageInstall {
         // present for backwards compatibility with older Habitat releases.
         env.remove(PATH_KEY);
 
-        let path = env::join_paths(self.runtime_paths()?)?
+        let mut paths = self.runtime_paths()?;
+
+        // Let's join the paths to the FS_ROOT
+        // In most cases, this does nothing and should only mutate
+        // the paths in a windows studio where FS_ROOT_PATH will
+        // be the studio root path (ie c:\hab\studios\...)
+        for path in &mut paths {
+            *path = fs::fs_rooted_path(&path, Some(&*self.fs_root_path));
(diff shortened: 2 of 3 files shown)


thesentinels:

Thank you for contributing!

components/core/src/package/install.rs
```diff
@@ -237,12 +237,20 @@ impl PackageInstall {
...
         env.remove(PATH_KEY);
 
-        let path = env::join_paths(self.runtime_paths()?)?
+        let mut paths = self.runtime_paths()?;
+
+        for path in &mut paths {
+            if path.starts_with("/") {
+                *path = Path::new(&*self.fs_root_path).join(path.strip_prefix("/").unwrap());
+            }
+        }
```
baumanj: Is this code appropriate for unix as well as windows? I feel like this deserves at least a comment to explain what's going on here.

components/core/src/package/install.rs
```diff
@@ -1420,13 +1428,24 @@ core/bar=pub:core/publish sub:core/subscribe
...
         expected.insert(
             "PATH".to_string(),
             env::join_paths(vec![
-                pkg_prefix_for(&pkg_install).join("bin"),
-                pkg_prefix_for(&other_pkg_install).join("sbin"),
+                fs_root_path.join(
+                    pkg_prefix_for(&pkg_install)
+                        .strip_prefix("/")
+                        .unwrap()
+                        .join("bin"),
```
baumanj: Since this is basically the same logic as above, maybe abstract it into a function? That would be a nice opportunity to give it a name, aiding readability.

components/core/src/package/install.rs
mwrock: well its effectively a noop on linux where `fs_root` is `/`. I'll copy the comments from the supervisor implementation that makes this explicit.

baumanj: While it's currently (and likely always will be) true, there's an implicit assumption in this code that [`fs_root_path()`](https://github.com/habitat-sh/core/blob/master/components/core/src/fs.rs#L408) will always be `/` on unix. If for some reason it weren't, this would be broken. Imagine if `fs_root_path()` returned `/foo`. Then `/bar` → `/foo/bar`, but `/foo/bar` → `/foo/foo/bar`.

I think there should probably be a check that `self.fs_root_path != "/"` before doing this and/or a `if !cfg(windows)`.

components/core/src/fs.rs
```diff
@@ -197,6 +197,26 @@ where
...
+{
+    match fs_root {
+        Some(root) => {
+            if path.starts_with("/") && cfg!(windows) {
+                Path::new(root.as_ref()).join(path.strip_prefix("/").unwrap())
+            } else {
+                path.to_path_buf()
+            }
+        }
+        None => path.to_path_buf(),
```
baumanj: You can eke out a smidgeon more concision:
```rust
        Some(ref root) if path.starts_with("/") && cfg!(windows) => {
            Path::new(root.as_ref()).join(path.strip_prefix("/").unwrap())
        }
        _ => path.to_path_buf(),
```

components/core/src/package/install.rs
```diff
@@ -237,12 +237,22 @@ impl PackageInstall {
         // present for backwards compatibility with older Habitat releases.
         env.remove(PATH_KEY);
 
-        let path = env::join_paths(self.runtime_paths()?)?
+        let mut paths = self.runtime_paths()?;
+
+        // Lets join the paths to the FS_ROOT
```
baumanj: Lets → Let's

mwrock:
@baumanj I think i have address your comments

PR #90, opened 2018-11-17
moving some common functionality from habitat

mwrock:
This PR accompanies https://github.com/habitat-sh/habitat/pull/5866 which moves most of the hook and templating code from the supervisor into the `common` crate of the habitat repo. There was some functionality that also needed to be used commonly but seemed better suited here:

```
sup/src/fs.rs -> folded into /core/src/fs.rs
sup/src/sys/abilities -> folded into /core/src/os/users
sup/src/sys/user -> folded into /core/src/os/users
sup/src/manager/service/dir.rs -> folded into /core/src/fs.rs
```

### DIFF
--- a/components/core/Cargo.toml
+++ b/components/core/Cargo.toml
@@ -37,6 +37,9 @@ url = "*"
 [target.'cfg(not(windows))'.dependencies]
 users = "*"
 
+[target.'cfg(target_os = "linux")'.dependencies]
+caps = "*"
+
 [target.'cfg(windows)'.dependencies]
 ctrlc = "*"
 habitat_win_users = { path = "../win-users" }
--- a/components/core/src/error.rs
+++ b/components/core/src/error.rs
@@ -135,7 +135,7 @@ pub enum Error {
     PackageUnpackFailed(String),
     /// When an error occurs parsing an integer.
     ParseIntError(num::ParseIntError),
-    /// Occurs when setting ownership or permissions on a file or directory fails.
+    /// Occurs upon errors related to file or directory permissions.
     PermissionFailed(String),
     /// Error parsing the contents of a plan file were incomplete or malformed.
     PlanMalformed,
@@ -458,7 +458,7 @@ impl error::Error for Error {
             Error::PackageNotFound(_) => "Cannot find a package",
             Error::PackageUnpackFailed(_) => "Package could not be unpacked",
             Error::ParseIntError(_) => "Failed to parse an integer from a string!",
-            Error::PermissionFailed(_) => "Failed to set permissions",
+            Error::PermissionFailed(_) => "File system permissions error",
             Error::PlanMalformed => "Failed to read or parse contents of Plan file",
             Error::PrivilegeNotHeld => "Privilege not held to spawn process as different user",
             Error::RegexParse(_) => "Failed to parse a regular expression",
--- a/components/core/src/os/users/linux.rs
+++ b/components/core/src/os/users/linux.rs
@@ -12,10 +12,28 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use caps::{self, CapSet, Capability};
 use std::path::PathBuf;
 
+use crate::error::{Error, Result};
 use crate::linux_users;
 use crate::linux_users::os::unix::{GroupExt, UserExt};
+
+/// This is currently the "master check" for whether the Supervisor
+/// can behave "as root".
+///
+/// All capabilities must be present. If we can run processes as other
+/// users, but can't change ownership, then the processes won't be
+/// able to access their files. Similar logic holds for the reverse.
+pub fn can_run_services_as_svc_user() -> bool {
+    has(Capability::CAP_SETUID) && has(Capability::CAP_SETGID) && has(Capability::CAP_CHOWN)
+}
+
+/// Helper function; does the current thread have `cap` in its
+/// effective capability set?
+fn has(cap: Capability) -> bool {
+    caps::has_cap(None, CapSet::Effective, cap).unwrap_or(false)
+}
 
 pub fn get_uid_by_name(owner: &str) -> Option<u32> {
     linux_users::get_user_by_name(owner).map(|u| u.uid())
@@ -66,4 +84,53 @@ pub fn get_home_for_user(username: &str) -> Option<PathBuf> {
 
 pub fn root_level_account() -> String {
     "root".to_string()
+}
+
+/// This function checks to see if a user and group and if:
+///     a) we are root
+///     b) we are the specified user:group
+///     c) fail otherwise
+pub fn assert_pkg_user_and_group(user: &str, group: &str) -> Result<()> {
+    if let None = get_uid_by_name(user) {
+        return Err(Error::PermissionFailed(format!(
+            "Package requires user {} to exist, but it doesn't",
+            user
+        )));
+    }
+    if let None = get_gid_by_name(&group) {
+        return Err(Error::PermissionFailed(format!(
+            "Package requires group {} to exist, but it doesn't",
+            group
+        )));
+    }
+
+    let current_user = get_current_username();
+    let current_group = get_current_groupname();
+
+    if let None = current_user {
+        return Err(Error::PermissionFailed(
+            "Can't determine current user".to_string(),
+        ));
+    }
+
+    if let None = current_group {
+        return Err(Error::PermissionFailed(
+            "Can't determine current group".to_string(),
+        ));
+    }
+
+    let current_user = current_user.unwrap();
+    let current_group = current_group.unwrap();
+
+    if current_user == root_level_account() {
+        Ok(())
+    } else {
+        if current_user == user && current_group == group {
+            // ok, sup is running as svc_user/svc_group already
+            Ok(())
+        } else {
+            let msg = format!("Package must run as {}:{} or root", user, &group);
+            return Err(Error::PermissionFailed(msg));
+        }
+    }
 }
--- a/components/core/src/os/users/windows.rs
+++ b/components/core/src/os/users/windows.rs
@@ -17,8 +17,14 @@ use std::path::PathBuf;
 
 use habitat_win_users::account::Account;
 
+use crate::error::{Error, Result};
+
 extern "C" {
     pub fn GetUserTokenStatus() -> u32;
+}
+
+pub fn can_run_services_as_svc_user() -> bool {
+    true
 }
 
 fn get_sid_by_name(name: &str) -> Option<String> {
@@ -64,6 +70,18 @@ pub fn root_level_account() -> String {
     env::var("COMPUTERNAME").unwrap().to_uppercase() + "$"
 }
 
+/// Windows does not have a concept of "group" in a Linux sense
+/// So we just validate the user
+pub fn assert_pkg_user_and_group(user: &str, _group: &str) -> Result<()> {
+    match get_uid_by_name(user) {
+        Some(_) => Ok(()),
+        None => Err(Error::PermissionFailed(format!(
+            "Package requires user {} to exist, but it doesn't",
+            user
+        ))),
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use std::env;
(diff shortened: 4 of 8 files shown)


thesentinels:

Thank you for contributing!

mwrock:
related: https://github.com/habitat-sh/habitat/issues/5161

baumanj:
Starting to review this now

components/core/src/error.rs
```diff
@@ -168,11 +179,17 @@ pub enum Error {
     WrongActivePackageTarget(package::PackageTarget, package::PackageTarget),
 }
 
+unsafe impl Send for Error {}
+unsafe impl Sync for Error {}
```
baumanj: I'm curious why these are necessary/appropriate. According to the docs, [`Send`](http://doc.rust-lang.org/1.30.1/core/marker/trait.Send.html) and [`Sync`](http://doc.rust-lang.org/1.30.1/core/marker/trait.Sync.html) are "automatically implemented when the compiler determines it's appropriate".

mwrock: It was the only way I could get of a compile error compiling the supervisor against this. Something releted to not implementing send. I don't remember it well. I'll dig it up.

components/core/src/templating/hooks.rs
```diff
@@ -0,0 +1,1591 @@
...
+        table
+    }
+
+    pub fn from_package_install(package: &PackageInstall, config_from: Option<&PathBuf>) -> Self {
+        Self::load(
+            &package.ident.name,
+            config_from
+                .and_then(|p| Some(p.as_path()))
+                .unwrap_or(&package.installed_path)
+                .join("hooks"),
```
baumanj: I think this can be simplified to:
```rust
            config_from.unwrap_or(&package.installed_path).join("hooks"),
```

baumanj: Also, the only caller I see for this function passes `None` for `config_from`, so it seems like we could eliminate that argument entirely and just pass
```rust
            package.installed_path.join("hooks"),
```
to `Self::load`. If that argument *is* needed, I'd rather eliminate the `Option` and make the callers do:
```rust
from_package_install(package, &package.installed_path)
// vs
from_package_install(package, &some_other_path)
```
Since I think that's more evocative to read at the call-site than
```rust
from_package_install(package, None)
// vs
from_package_install(package, &some_other_path)
```

mwrock: The `config_from` is explicitly passed by the supervisor. You just don't see it since that is in a separate repo.

components/core/src/error.rs
mwrock: heh...go figure...I can't repro that error now. I have a hunch that is due to some changes I made yesterday in the supervisor.

components/core/Cargo.toml
```diff
@@ -10,11 +10,13 @@ base64 = "*"
 gcc = "0.3"
 
 [dependencies]
+actix-web = { version = "*", default-features = false }
```
christophermaier: Why do we need Actix in core now?

components/core/src/error.rs
```diff
@@ -37,6 +37,7 @@ pub enum Error {
     /// Occurs when a `habitat_core::package::PackageArchive` is being read.
     ArchiveError(libarchive::error::ArchiveError),
     BadBindingMode(String),
+    BadEnvConfig(String),
```
christophermaier: It'd be nice to have a doc comment on this, similar to the others.

components/core/src/error.rs
```diff
@@ -135,18 +136,28 @@ pub enum Error {
...
-    /// Occurs when setting ownership or permissions on a file or directory fails.
+    /// Occurs upon errors related to file or directory permissions.
     PermissionFailed(String),
     /// Error parsing the contents of a plan file were incomplete or malformed.
     PlanMalformed,
     // When CreateProcessAsUserW does not have the correct privileges
     PrivilegeNotHeld,
     /// When an error occurs parsing or compiling a regular expression.
     RegexParse(regex::Error),
+    /// Whwn an error occurs serializing rendering context
```
christophermaier: "Whwn" -> "When"

components/core/src/templating/health.rs
```diff
@@ -0,0 +1,89 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::fmt;
+
+use actix_web::http::StatusCode;
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize)]
+pub enum HealthCheck {
```
christophermaier: Does this really need to be over here? What would anything outside the Supervisor need this for?

components/core/src/templating/hooks.rs
baumanj: Yeah, I see [here](https://github.com/habitat-sh/habitat/blob/install_hook/components/sup/src/manager/service/mod.rs#L167) and [here](https://github.com/habitat-sh/habitat/blob/install_hook/components/sup/src/manager/service/mod.rs#L619).

I think I get it now. This is an `Option` because [`ServiceSpec`'s field is an option](https://github.com/habitat-sh/habitat/blob/install_hook/components/sup/src/manager/service/spec.rs#L272). That makes more sense.

I think we may be better served by removing the optionality from `ServiceSpec`; I'm not convinced it's necessary. But that's a change for another day.

components/core/src/templating/mod.rs
```diff
@@ -0,0 +1,414 @@
...
+    let pkg = package::Pkg::from_install(package.clone())?;
+
+    fs::SvcDir::new(&pkg.name, &pkg.svc_user, &pkg.svc_group).create()?;
+
+    let cfg = config::Cfg::new(&pkg, None)?;
+    let ctx = RenderContext::new(&pkg, &cfg);
+    let cfg_renderer = config::CfgRenderer::new(pkg.path.join("config"))?;
+    cfg_renderer.compile(&pkg.name, &pkg, &ctx)?;
+
+    hooks::HookTable::from_package_install(package).compile(&pkg.name, &ctx);
```
christophermaier: What happens if you try to compile a hook that depends on things that aren't present in this limited rendering context (like census information)?

Seems like we should throw an error.

components/core/src/templating/test_helpers.rs
```diff
@@ -0,0 +1,106 @@
+// Copyright (c) 2017 Chef Software Inc. and/or applicable contributors
```
christophermaier: These also got moved to https://github.com/habitat-sh/habitat/blob/master/components/sup/src/test_helpers.rs... if we're moving them over here, the other tests in the supervisor that use them should pull them from here.

components/core/Cargo.toml
mwrock: `HealthCheck` uses its `StatusCode` but based on your point below, we probably don't need that here after all.

components/core/src/templating/mod.rs
mwrock: I'm assuming this comment is also targetted to config templates also? Currently it just renders nothing if there is nothing in the context. Thinking of the following example scenario, I'd prefer to raise a warning instead of an error:

There may be multiple config templates and some have, for example, binds and others do not. When the `install` hook fires at install time, it will need to compile all templates since it cannot assume which templates it will use. We wouldn't want the templates with the binds to error during installation. 

Also for debug purposes, one might use `hab pkg compile` and there would be no `bind` context. So a warning informing the cli user that there is missing context would be nice but they could still render the templates.

mwrock: One extra point. I'm not certain yet how much we can control for that via the serde API.

components/core/src/templating/health.rs
mwrock: hmm. Keeping this in the supervisor actually looks a bit messy. It is referenced several times in the `HealthCheckHook`. Given how the `HookTable` works, all hooks need to be defined in core.

components/core/Cargo.toml
mwrock: Its used by `HealthCheck`. I looked at moving that back to the supervisor as you suggested below but its pretty well coupled with the hooks (see my comment below).

components/core/src/templating/mod.rs
christophermaier: I also thought we were going to have a specific install hook, or am I misremembering?

That might end up being cleaner, since we wouldn't necessarily need to bring over all the hooks, just the handlebars and context stuff.

mwrock: Yeah there is an `InstallHook` added to `hooks.rs`. Maybe it makes better sense for the `HookTable` not to include that hook and then just keep `HookTable` and the non-install hook implementations with the supervisor.

components/core/src/templating/test_helpers.rs
mwrock: ends up that the only files using these functions were moved over

components/core/src/error.rs
mwrock: Not sure why github is not showing this as "outdated" its been fixed

components/core/src/templating/mod.rs
mwrock: The downside of keeping the non-`install` hooks with the supervisor is that the `hab pkg compile` cli command will not be able to compile them. Also, future work is going to need to be able to compile the health-check via the cli.

components/core/src/error.rs
baumanj: The issue here is the [addition of `handlebars::error::RenderError` to the core Error enum](https://github.com/habitat-sh/core/blob/aedded85c1325b5f4f153e6cc8bc39bb7cabe9c4/components/core/src/error.rs#L157). The version we're at fails to require `Sync` in one of its fields. See https://github.com/sunng87/handlebars-rust/issues/194.

mwrock:
@christophermaier @baumanj  Still not great but what do you think of https://github.com/habitat-sh/core/pull/90/commits/f320dffc8b2d2adeb637f8886847300d8df0c84b for dealing with the `RenderError` issues?

baumanj:
> @christophermaier @baumanj Still not great but what do you think of [f320dff](https://github.com/habitat-sh/core/commit/f320dffc8b2d2adeb637f8886847300d8df0c84b) for dealing with the `RenderError` issues?

I think that's good! It's clean and small and I don't think should negatively impact the utility of the error messages at all. We might want to add a note to change it back when we upgrade the handlebars crate (maybe an issue, linked from our comments at the pinning in Cargo.toml), but otherwise I'm happy with it.

components/core/src/os/users/linux.rs
```diff
@@ -12,11 +12,37 @@
...
+/// able to access their files. Similar logic holds for the reverse.
+#[cfg(target_os = "linux")]
+pub fn can_run_services_as_svc_user() -> bool {
+    has(Capability::CAP_SETUID) && has(Capability::CAP_SETGID) && has(Capability::CAP_CHOWN)
+}
+
+#[cfg(target_os = "macos")]
+pub fn can_run_services_as_svc_user() -> bool {
+    true
+}
```
christophermaier: We should move this out of this module, since it doesn't pertain to Linux.

I'm not sure we really even need this implementation; I don't think any macOS code would make it down to anyplace that this function would be needed.

In the same spirit, if we're only pulling this module in on Linux, then all the `#[cfg(target_os = "linux")]` annotations end up being noise.

components/core/src/os/users/windows.rs
```diff
@@ -72,13 +90,17 @@ mod tests {
...
+        env::set_var("USERNAME", orig_user);
     }
 
     #[test]
     fn return_none_when_no_user() {
+        let orig_user = get_current_username().unwrap();
         env::remove_var("USERNAME");
-        assert_eq!(get_current_username(), None)
+        assert_eq!(get_current_username(), None);
+        env::set_var("USERNAME", orig_user);
```
christophermaier: Any chance these two tests end up stepping on each other? They're both manipulating global state.

components/core/src/os/users/linux.rs
mwrock: yup. I just double checked and don't see any scenario where code would reach here on a mac until it runs a supervisor and we support darwin package installs.

components/core/src/os/users/windows.rs
mwrock: TIL: `cargo check` runs tests in parallel. So yes.

I'd prefer to tackle this separately by refactoring the windows implementation of `get_current_username` to call the `GetUserNameW` API rather than checking the environment variable. Will file an issue.

christophermaier: Using an API call is probably a better way forward. And yeah, it sounds worth a separate PR 👍 

If you do end up sticking with the environment variable approach, though, you'll want to look at using our new-ish [locked env var](https://github.com/habitat-sh/habitat/blob/master/components/common/src/locked_env_var.rs) tooling (which would also  mean pulling it over here to the core crate).

PR #97, opened 2018-12-19
Add trait for converting numeric types to i64

raskchanky:
This is necessary for https://github.com/habitat-sh/habitat/pull/5986

Signed-off-by: Josh Black <raskchanky@gmail.com>

### DIFF
--- a/components/core/src/util/mod.rs
+++ b/components/core/src/util/mod.rs
@@ -80,11 +80,15 @@ trait ToI64 {
 impl ToI64 for usize {
     fn to_i64(self) -> i64 {
         if mem::size_of::<usize>() >= mem::size_of::<i64>() && self > i64::max_value() as usize {
-            error!(
-                "Tried to convert an out-of-range usize ({}) to i64; using i64::max_value()",
-                self
-            );
-            i64::max_value()
+            if cfg!(debug_assertions) {
+                panic!("Tried to convert an out-of-range usize ({}) to i64", self);
+            } else {
+                error!(
+                    "Tried to convert an out-of-range usize ({}) to i64; using i64::max_value()",
+                    self
+                );
+                i64::max_value()
+            }
         } else {
             self as i64
         }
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/core/src/util/mod.rs
```diff
@@ -70,3 +71,63 @@ where
...
+}
+
+impl ToI64 for usize {
+    fn to_i64(self) -> i64 {
+        if mem::size_of::<usize>() >= mem::size_of::<i64>() && self > i64::max_value() as usize {
+            error!(
+                "Tried to convert an out-of-range usize ({}) to i64; using i64::max_value()",
+                self
+            );
+            i64::max_value()
```
raskchanky: @baumanj I left out the `panic!` from your original suggestion.  This felt more ergonomic to me.

Also, I could be wrong, but I know at one point our release builds of hab were not proper release builds, but were actually debug builds, so I would hate to have panics happen for our users in the unlikely event that this branch gets triggered.

baumanj: I agree that we should avoid panicking in this case in release, but I think our chances of catching an issue caused by a bad conversion go way down if we don't panic in debug mode. I'm pretty sure our releases are proper releases and should not include `cfg!(debug_assertions)` code, but let's figure that out for sure.

As we add more code that does things like collect metrics, the value of being able to panic only in debug builds goes way up.

components/core/src/util/mod.rs
```diff
@@ -70,3 +71,63 @@ where
...
+        } else {
+            self as i64
+        }
+    }
+}
+
+impl ToI64 for u64 {
+    fn to_i64(self) -> i64 {
+        if mem::size_of::<u64>() >= mem::size_of::<i64>() && self > i64::max_value() as u64 {
```
baumanj: There's no need to test the relative sizes of `u64` and `i64`; unlike `usize`, [they're intrinsic to the type](http://doc.rust-lang.org/1.31.0/std/mem/fn.size_of.html).
```suggestion
        if self > i64::max_value() as u64 {
```

raskchanky: Ah, of course.  I just copy/pasted and spaced on that.

components/core/src/util/mod.rs
raskchanky: Ok, fair enough.

PR #107, opened 2019-01-18
Replace Strings and &strs with type-safe habitat_core::ChannelIdent type

baumanj:
The potential for using `Default::default()` instead of `channel::default()` when the type representing a channel is a `&str` or `String` is a nasty latent error that the compiler will not find. To avoid it, we should use a proper type to represent channel names.

There are corresponding changes to the [habitat](https://github.com/habitat-sh/habitat/pull/6074) and [builder](https://github.com/habitat-sh/builder/pull/881) repos that must be merged after this. After that, [channel.rs](https://github.com/habitat-sh/core/blob/master/components/core/src/channel.rs) can be removed entirely.

### DIFF
--- a/components/core/src/lib.rs
+++ b/components/core/src/lib.rs
@@ -90,10 +90,6 @@ impl ChannelIdent {
         crate::env::var(key).map(ChannelIdent)
     }
 
-    pub fn from(s: &str) -> Self {
-        ChannelIdent(s.to_string())
-    }
-
     pub fn as_str(&self) -> &str {
         self.0.as_str()
     }
@@ -118,6 +114,18 @@ impl ChannelIdent {
     /// Helper function for Builder dynamic channels
     pub fn bldr_name(id: u64) -> String {
         format!("bldr-{}", id)
+    }
+}
+
+impl From<&str> for ChannelIdent {
+    fn from(s: &str) -> Self {
+        ChannelIdent(s.to_string())
+    }
+}
+
+impl From<String> for ChannelIdent {
+    fn from(s: String) -> Self {
+        ChannelIdent(s)
     }
 }
 
--- a/components/core/src/lib.rs
+++ b/components/core/src/lib.rs
@@ -55,6 +55,9 @@ pub mod service;
 pub mod url;
 pub mod util;
 
+use std::env::VarError;
+use std::ffi::OsStr;
+use std::fmt;
 use std::path::PathBuf;
 
 pub use crate::os::filesystem;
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/core/src/lib.rs
```diff
@@ -68,3 +71,63 @@ lazy_static::lazy_static! {
...
+pub struct ChannelIdent(String);
+
+impl ChannelIdent {
+    const LEGACY_ENVVAR: &'static str = "HAB_DEPOT_CHANNEL";
+    pub const BLDR_ENVVAR: &'static str = "HAB_BLDR_CHANNEL";
+    const UNSTABLE: &'static str = "unstable";
+    const STABLE: &'static str = "stable";
+
+    pub fn from_env_var(key: impl AsRef<OsStr>) -> std::result::Result<Self, VarError> {
+        crate::env::var(key).map(ChannelIdent)
```
christophermaier: That's pretty slick

components/core/src/lib.rs
```diff
@@ -68,3 +71,63 @@ lazy_static::lazy_static! {
...
+        write!(f, "{}", self.0)
+    }
+}
+
+impl Default for ChannelIdent {
+    fn default() -> Self {
+        match env::var(ChannelIdent::BLDR_ENVVAR) {
+            Ok(value) => ChannelIdent(value.to_string()),
+            Err(_) => ChannelIdent::legacy_default(),
+        }
```
christophermaier: Based on some awkward situations I've seen with other of our `Default` implementations that used environment variables, I'm a little suspicious of this, since it conflates a default value with something that could could be configured to be something else. In particular, `HAB_BLDR_CHANNEL` doesn't define a default value, but is a way for the user to specify that they want to operate on something _other than the default channel_.

This is one of the things I like about the `EnvConfig` abstraction, because these two cases are clearly distinguished. Sometimes we want the stable channel, and sometimes we want to know what the user wants. Often those are the same, but sometimes they're not.

components/core/src/lib.rs
```diff
@@ -68,3 +71,63 @@ lazy_static::lazy_static! {
...
+
+/// A Builder channel
+#[derive(Clone, Debug, PartialEq)]
+pub struct ChannelIdent(String);
+
+impl ChannelIdent {
+    const LEGACY_ENVVAR: &'static str = "HAB_DEPOT_CHANNEL";
+    pub const BLDR_ENVVAR: &'static str = "HAB_BLDR_CHANNEL";
+    const UNSTABLE: &'static str = "unstable";
+    const STABLE: &'static str = "stable";
```
christophermaier: :heart: embedded constants

components/core/src/lib.rs
```diff
@@ -68,3 +71,63 @@ lazy_static::lazy_static! {
             .to_string()
     };
 }
+
+/// A Builder channel
+#[derive(Clone, Debug, PartialEq)]
+pub struct ChannelIdent(String);
```
christophermaier: What do you feel are the benefits of `ChannelIdent` over `Channel`?

To me, the former is clearly more explicit, but in the absence of a distinct existing `Channel` type (similar to what we have with packages and package identifiers), it reads more verbosely than `Channel` alone.

components/core/src/lib.rs
```diff
@@ -68,3 +71,63 @@ lazy_static::lazy_static! {
...
+        ChannelIdent(
+            env::var(Self::LEGACY_ENVVAR)
+                .ok()
+                .and_then(|c| Some(c.to_string()))
+                .unwrap_or(Self::STABLE.to_string()),
+        )
+    }
+
+    /// Helper function for Builder dynamic channels
+    pub fn bldr_name(id: u64) -> String {
```
christophermaier: Might rename this to something like `sandbox_channel`, `bldr_sandbox_channel`, etc., since "sandbox" is part of our "ubiquitous language".

Might also be worth just returning this as a `ChannelIdent` directly, rather than a `String`. That would discourage us from lapsing back to stringly-typed APIs. If you need just the name, you can still get it from the `ChannelIdent` API, and there's no additional overhead (`ChannelIdent` is a zero-cost wrapper, and you already have to create a `String` here anyway).

components/core/src/lib.rs
```diff
@@ -68,3 +71,63 @@ lazy_static::lazy_static! {
...
+    const UNSTABLE: &'static str = "unstable";
+    const STABLE: &'static str = "stable";
+
+    pub fn from_env_var(key: impl AsRef<OsStr>) -> std::result::Result<Self, VarError> {
+        crate::env::var(key).map(ChannelIdent)
+    }
+
+    pub fn from(s: &str) -> Self {
+        ChannelIdent(s.to_string())
+    }
```
christophermaier: Why not implement `From<&str>` (or even `From<String>`)?

components/core/src/lib.rs
christophermaier: Oh, that certainly does change things 😄 

`ChannelIdent` it is!

components/core/src/lib.rs
baumanj: Yeah, that would probably be better. Will update!

components/core/src/lib.rs
baumanj: This is only used one place in builder, and it's getting passed directly in to protobuf code, so I think we might as well just use a `String`. That said, this was just an effort to keep the interface the same and now that I see it's only used one place, I think we can just do away with this function altogether and inline the code.

components/core/src/lib.rs
baumanj: I'm all for making this use our `EnvConfig` abstraction, but I'd like to do it in a follow-on PR since this PR isn't trying to change any functionality, just refactor to a more type-safe implementation of the existing behavior.

components/core/src/lib.rs
baumanj: Addressed [here](https://github.com/habitat-sh/core/pull/107/commits/eb44c29b83bd2431e876d547c2fddab106278226).

See the corresponding improvements to the [`habitat`](https://github.com/habitat-sh/habitat/pull/6074/commits/4ec82604a9718304fad72912ef9bb147b4165f60) and [`builder`](https://github.com/habitat-sh/builder/pull/881/commits/947548e84cf05c8c9af63e9ebf5d26d55bf99199) PRs too.

components/core/src/lib.rs
baumanj: This is [removed](https://github.com/habitat-sh/core/pull/107/commits/e88095b78a983f504ceebe2e6cfb6dc815541409) and [inlined in `builder`](https://github.com/habitat-sh/builder/pull/881/commits/6adf3fedb27ddc04641dcd4f9a20ad2912391955) now.

baumanj:
I think all your comments have been addressed now, @christophermaier.

PR #109, opened 2019-01-24
Enable core Clippy lints

christophermaier:
This brings the Clippy lints being used in line with what we're currently using in [habitat-sh/habitat](https://github.com/habitat-sh/habitat).

Keeping them in sync helps when developing features that involve both crates.

### DIFF
--- a/components/core/src/crypto/keys/sym_key.rs
+++ b/components/core/src/crypto/keys/sym_key.rs
@@ -341,7 +341,6 @@ impl SymKey {
                         val
                     )));
                 }
-                ()
             }
             None => {
                 let msg = format!(
--- a/components/core/src/crypto/keys/sym_key.rs
+++ b/components/core/src/crypto/keys/sym_key.rs
@@ -620,7 +620,7 @@ mod test {
         let pair = SymKey::generate_pair_for_ring("beyonce").unwrap();
         pair.to_pair_files(cache.path()).unwrap();
 
-        let (nonce, ciphertext) = pair.encrypt("Ringonit".as_bytes()).unwrap();
+        let (nonce, ciphertext) = pair.encrypt(b"Ringonit").unwrap();
         let message = pair.decrypt(&nonce, &ciphertext).unwrap();
         assert_eq!(message, "Ringonit".to_string().into_bytes());
     }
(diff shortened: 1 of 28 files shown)


thesentinels:

Thank you for contributing!

components/core/src/crypto/keys/sym_key.rs
```diff
@@ -71,28 +71,24 @@ impl SymKey {
...
                 debug!(
                     "Can't find secret key for name_with_rev {}: {}",
                     name_with_rev, e
                 );
-                None
-            }
-        };
+                e
+            })
+            .ok();
```
christophermaier: The lint was flagging the `Ok(k) => Some(k)` lines for some reason. Rather than disabling it here, however, I felt this refactoring was a bit more clear.

christophermaier:
Merged, despite the DCO bot being uncooperative; all commits have the appropriate sign-off.

components/core/src/crypto/keys/sym_key.rs
baumanj: The issue was https://rust-lang.github.io/rust-clippy/master/#unit_arg because when constructing `Some(k)` in the `pk` (public key) case, `k` was the unit value `()`. This is because `SymKey` is just an alias for a particular monomorphization of the generic type `KeyPair<P, S>`: `KeyPair<(), SymSecretKey>`.

Given that, passing `KeyPair::new(name: String, rev: String, p: Option<P>, s: Option<S>)` a `Some(())` value for the public key, seems kind of silly. Wouldn't it be clearer to eliminate the whole block for setting `pk` to a value that will never be relevant and instead call `Self::new` with:
```rust
        Ok(Self::new(name, rev, None, sk))
```
?

This further allows us to eliminate the unnecessary (and frankly misleading) implementation of `SymKey::get_public_key` and reduces `SymKey::get_pair_for` to
```rust
    pub fn get_pair_for<P: AsRef<Path> + ?Sized>(
        name_with_rev: &str,
        cache_key_path: &P,
    ) -> Result<Self> {
        let (name, rev) = parse_name_with_rev(&name_with_rev)?;
        let sk = match Self::get_secret_key(name_with_rev, cache_key_path.as_ref()) {
            Ok(k) => Some(k),
            Err(e) => {
                let msg = format!(
                    "No secret keys found for name_with_rev {}: {}",
                    name_with_rev, e
                );
                return Err(Error::CryptoError(msg));
            }
        };
        Ok(Self::new(name, rev, None, sk))
    }
```

christophermaier:
@baumanj Yeah, that sounds like a great thing to change. I'll take a look at that right now.

PR #110, opened 2019-01-28
SymKeys don't have public keys, so don't pretend they do

christophermaier:
SymKeys are currently declared to have a public key type of `()` due
to their being a monomorphization of the `KeyPair<P,S>` type. When
creating a new `SymKey`, you can either pass a "public key" of
`Some(())` or `None`, which is arguably more correct (given the
existing `KeyPair` typing, that is... addressing that would be a
larger refactoring).

By embracing the `None`, we get to eliminate the (private)
`get_public_key` method (which is very confusing in this context), as
well as greatly simplify the implementation of `get_pair_for`.

Hat tip to @baumanj for this!

This commit contains a partial reversion of commit
971f5efb3b65f716eba062937321e791fb7f11b2.

Signed-off-by: Christopher Maier <cmaier@chef.io>

### DIFF
--- a/components/core/src/crypto/keys/sym_key.rs
+++ b/components/core/src/crypto/keys/sym_key.rs
@@ -71,32 +71,17 @@ impl SymKey {
         cache_key_path: &P,
     ) -> Result<Self> {
         let (name, rev) = parse_name_with_rev(&name_with_rev)?;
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

PR #112, opened 2019-01-29
Derive necessary traits for ChannelIdent: Deserialize, Serialize, Eq, Hash

baumanj:
Further work on integrating `ChannelIdent` into [habitat](https://github.com/habitat-sh/habitat/pull/6074) and [builder](https://github.com/habitat-sh/builder/pull/881) showed the need for this additional functionality. Those PRs will be blocked until this merges.

### DIFF
--- a/components/core/src/lib.rs
+++ b/components/core/src/lib.rs
@@ -77,7 +77,7 @@ lazy_static::lazy_static! {
 }
 
 /// A Builder channel
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Deserialize, Serialize, Clone, Debug, Eq, Hash, PartialEq)]
 pub struct ChannelIdent(String);
 
 impl ChannelIdent {


thesentinels:

Thank you for contributing!

components/core/src/lib.rs
```diff
@@ -77,7 +77,7 @@ lazy_static::lazy_static! {
 }
 
 /// A Builder channel
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Deserialize, Serialize, Clone, Debug, Eq, Hash, PartialEq)]
```
christophermaier: Refresh my memory... will Serde serialize a `ChannelIdent` to and from a bare String (i.e., just the name), or is there some additional structure to reflect it's "newtype-ness"?

christophermaier: If the former, then :shipit:

baumanj: I was worried about that as well, so I tested it in both the HTTP gateway's services output and the generated .spec file: it indeed generates only a bare string, and [here's a playground that demonstrates it in minimal form](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f8293a0d7ce2c140710da7be89dac448).

PR #114, opened 2019-01-31
[core] Remove Linux capabilities code from macOS target.

fnichol:
This change allows the `core` crate to be compiled on macOS systems and additionally eliminates almost all of the `extern crate` declarations which cuts down on the number platform-dependent attributes.

I have some further refactoring work to try and clarify the differences we have between "unixlike" systems (i.e. Linux and macOS) and Windows systems. I'm hoping that this changeset is small enough to fix the macOS target and reduce short term confusion until we clean up a bit more. :cake:

### DIFF
--- a/components/core/src/os/users/linux.rs
+++ b/components/core/src/os/users/linux.rs
@@ -15,8 +15,8 @@
 use std::path::PathBuf;
 
 use crate::error::{Error, Result};
-use crate::linux_users;
-use crate::linux_users::os::unix::{GroupExt, UserExt};
+use users;
+use users::os::unix::{GroupExt, UserExt};
 
 /// This is currently the "master check" for whether the Supervisor
 /// can behave "as root".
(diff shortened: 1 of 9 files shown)


components/core/src/os/users/linux.rs
```diff
@@ -12,40 +12,45 @@
...
 // limitations under the License.
 
-use caps::{self, CapSet, Capability};
 use std::path::PathBuf;
 
 use crate::error::{Error, Result};
-use crate::linux_users;
-use crate::linux_users::os::unix::{GroupExt, UserExt};
+use users;
+use users::os::unix::{GroupExt, UserExt};
```
christophermaier: While we're fixing things up, this file should be renamed to `unix.rs` for consistency.

fnichol:
Yeah, to both @christophermaier and @baumanj's points, I fully intend on renaming that module to the common `unix` family, extract the Linux-ness into a `linux` module, and likely extract the capability behavior into its own module of concern. I was hoping to get ourselves correctly as minimally as possible and then continue down the path as I'm waiting for CIs to compile/test/etc. 😄

components/core/src/os/users/linux.rs
fnichol: Totally, it pops right out once you take a second to look, right?

components/core/src/os/users/linux.rs
```diff
@@ -12,40 +12,45 @@
...
+    }
+
     has(Capability::CAP_SETUID) && has(Capability::CAP_SETGID) && has(Capability::CAP_CHOWN)
 }
 
-/// Helper function; does the current thread have `cap` in its
-/// effective capability set?
-fn has(cap: Capability) -> bool {
-    caps::has_cap(None, CapSet::Effective, cap).unwrap_or(false)
+#[cfg(target_os = "macos")]
```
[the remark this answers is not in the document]
fnichol: Yes, I completely agree and am itching to finish this up--hopefully we can be left with a better scaffolding (i.e. modules, naming, etc) that make these mistakes harder in the future.

Issue #15, opened 2018-03-09
BoxKeyPair::encrypt functions return type is misleading


baumanj:
The currently return type is `Result<Vec<u8>>`. This implies that the encrypted values are raw bytes, when in-fact they are bytes that represent a base64 encoded utf8 string. Either the return type should be changed to `Result<String>`  (and comments added to indicate that `String` is base64 encoded) or the return value should be raw bytes, not base64-encoded.

This caused [quite a bit of confusion](https://github.com/habitat-sh/habitat/pull/4706#pullrequestreview-101661074).

PR #115, opened 2019-02-04
Fix BoxKeyPair::encrypt return type

baumanj:
See https://github.com/habitat-sh/core/issues/15 for background.

The three commits represent three different levels of fix to `encrypt`'s return Result:
1. Change  from `Vec<u8>` to `String`
2. Use a new `WrappedSealedBox(String)` to improve safety, but add some copy overhead for some consumers
3. Make WrappedSealedBox a [copy-on-write](http://doc.rust-lang.org/1.32.0/std/borrow/enum.Cow.html) type for optimal type-safety *and* no-copy semantics, at the cost of some additional complexity.

Feedback on which approach is best is welcome.

Proposed changes for [`habitat`](https://github.com/habitat-sh/habitat/pull/6116) and [`builder`](https://github.com/habitat-sh/builder/pull/910) follow the same structure (though 2, and 3 for builder are identical).

### DIFF
--- a/components/core/src/crypto/keys/box_key_pair.rs
+++ b/components/core/src/crypto/keys/box_key_pair.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::borrow::Cow;
 use std::path::{Path, PathBuf};
 use std::str;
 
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

baumanj:
> I like the completeness of the `Cow`-based `WrappedSealedBox` approach. 👍

For what it's worth, I also looked into whether `ChannelIdent` would benefit from a similar approach and decided not. I also think the value is much less there since a channel should have a pretty short name where as the size of the encrypted box is essentially unbounded.

christophermaier:
@baumanj Yup, makes sense: only add complexity when it's worthwhile to do so.

baumanj:
Rebased

PR #116, opened 2019-02-05
Move ServiceBind implementation from Supervisor

jamesc:
This allow us to clean up some duplication code between `sup`,
`sup-protocol` and the k8s exporter around `ServiceBind`s

Once merged I've a commit on `habitat-sh/habitat` to remove the duplication

Signed-off-by: James Casey <james@chef.io>

### DIFF
--- a/components/core/src/error.rs
+++ b/components/core/src/error.rs
@@ -96,6 +96,8 @@ pub enum Error {
     FullyQualifiedPackageIdentRequired(String),
     /// Occurs when an application environment string cannot be successfully parsed.
     InvalidApplicationEnvironment(String),
+    /// Occurs when a service binding cannot be successfully parsed.
+    InvalidBinding(String),
     /// Occurs when a package identifier string cannot be successfully parsed.
     InvalidPackageIdent(String),
     /// Occurs when a package target string cannot be successfully parsed.
@@ -267,6 +269,11 @@ impl fmt::Display for Error {
                  is in the form application.environment (example: twitter.prod)",
                 e
             ),
+            Error::InvalidBinding(ref binding) => format!(
+                "Invalid binding \"{}\", must be of the form <NAME>:<SERVICE_GROUP> where <NAME> \
+                 is a service name, and <SERVICE_GROUP> is a valid service group",
+                binding
+            ),
             Error::InvalidPackageIdent(ref e) => format!(
                 "Invalid package identifier: {:?}. A valid identifier is in the form \
                  origin/name (example: acme/redis)",
@@ -430,6 +437,8 @@ impl error::Error for Error {
                 "Application environment strings must be in \
                  application.environment format (example: twitter.prod)"
             }
+            Error::InvalidBinding(_) => "Service Bind strings must be in \
+                name:service_group format (example cache:redis.cache@organization).",
             Error::InvalidPackageIdent(_) => {
                 "Package identifiers must be in origin/name format (example: acme/redis)"
             }
--- a/components/core/src/service.rs
+++ b/components/core/src/service.rs
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::error::{Error, Result};
+use regex::Regex;
+use serde_derive::{Deserialize, Serialize};
 use std::cmp::{Ordering, PartialOrd};
 use std::fmt;
 use std::num::ParseIntError;
@@ -19,11 +22,6 @@ use std::ops::{Deref, DerefMut};
 use std::result;
 use std::str::FromStr;
 use std::time::Duration;
-
-use regex::Regex;
-use serde_derive::{Deserialize, Serialize};
-
-use crate::error::{Error, Result};
 
 lazy_static::lazy_static! {
     static ref SG_FROM_STR_RE: Regex =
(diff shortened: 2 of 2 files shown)


thesentinels:

Thank you for contributing!

components/core/src/service.rs
```diff
@@ -499,6 +565,108 @@ mod test {
...
+
+        assert_eq!(bind.name, String::from("name"));
+        assert_eq!(
+            bind.service_group,
+            ServiceGroup::from_str("service.group").unwrap()
+        );
+    }
+
+    #[test]
+    #[should_panic(expected = "uhoh")]
```
christophermaier: I know these tests were copied over, but it might be better to assert on the error in the test, rather than just unwrapping and relying on `should_panic`. There's nothing inherent to the operation of `from_str` that is causing the panic; it's the `unwrap` call after that that is causing it.

components/core/src/error.rs
```diff
@@ -267,6 +269,11 @@ impl fmt::Display for Error {
                  is in the form application.environment (example: twitter.prod)",
                 e
             ),
+            Error::InvalidBinding(ref binding) => format!(
+                "Invalid binding \"{}\", must be of the form <NAME>:<SERVICE_GROUP> where <NAME> \
```
baumanj:
```suggestion
                "Invalid binding '{}', must be of the form <NAME>:<SERVICE_GROUP> where <NAME> \
```
We use single quotes for internal quoting [elsewhere](https://github.com/habitat-sh/core/blob/master/components/core/src/error.rs#L175) in [this file](https://github.com/habitat-sh/core/blob/master/components/core/src/error.rs#L346-L347), and since I find that easier to read than the escapes, I suggest we go that way.

components/core/src/service.rs
```diff
@@ -499,6 +565,108 @@ mod test {
...
+        let bind = ServiceBind {
+            name: String::from("name"),
+            service_group: ServiceGroup::from_str("service.group").unwrap(),
+        };
+
+        assert_eq!("name:service.group", bind.to_string());
+    }
+
+    #[test]
+    fn service_bind_toml_deserialize() {
```
christophermaier: Again, I know these were copied over, but I'm not really seeing the benefit of these tests. If we know that our `Serialize` and `Deserialize` implementations behave correctly, then ensuring that we can round-trip to TOML and back seems more like we're testing Serde at that point.

components/core/src/service.rs
```diff
@@ -83,7 +81,75 @@ impl FromStr for BindingMode {
     }
 }
 
-////////////////////////////////////////////////////////////////////////
+/// A binding from a service name to a service group that provides that service
+#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+pub struct ServiceBind {
+    pub name: String,
+    pub service_group: ServiceGroup,
```
baumanj: Can we make these not public? They're only references a couple places as far as I can tell.

components/core/src/service.rs
```diff
@@ -83,7 +81,75 @@ impl FromStr for BindingMode {
...
+    fn from_str(bind_str: &str) -> result::Result<Self, Self::Err> {
+        let values: Vec<&str> = bind_str.split(':').collect();
+        if values.len() == 2 {
+            Ok(ServiceBind {
+                name: values[0].to_string(),
+                service_group: ServiceGroup::from_str(values[1])?,
+            })
+        } else {
+            Err(Error::InvalidBinding(bind_str.to_string()))
+        }
```
baumanj: I'm really averse to mucking around with array indices and such when we have more expressive constructs available:
```rust
        match bind_str.split(':').collect::<Vec<_>>().as_slice() {
            [name, service_group] => Ok(ServiceBind::new(name, service_group)?),
            _ => Err(Error::InvalidBinding(bind_str.to_string())),
        }
```
This requires the addition of a `new` associated function, but I think that's a useful thing to have anyway:
```rust
impl ServiceBind {
    fn new(name: &str, service_group: &str) -> Result<Self> {
        Ok(Self {
            name: name.to_string(),
            service_group: ServiceGroup::from_str(service_group)?,
        })
    }
}
```

components/core/src/service.rs
jamesc: I had looked at making them private.  The only place I found post-refactor of hab was in the prost type conversion functions. I'd need to add some accessors to cover that case and wasn't sure if that was idiomatic - googling led me to https://www.reddit.com/r/rust/comments/4bs50v/are_accessors_idiomatic_in_rust/d1bvx23 which made me move back to the pub struct members.

Do you tend to never make struct fields `pub` ?

components/core/src/service.rs
jamesc: +1. This is better

components/core/src/service.rs
christophermaier: As with most things, "it depends", but a lot of entanglement we have in the Supervisor codebase can be traced back to unnecessary `pub` usage. This applies to struct fields, functions, modules, etc.

I generally make things private until the code tells me it really needs to be public.

components/core/src/service.rs
jamesc: I agree on the `Serialize` test as our trait implementation just calls through to Serde.  I think there is value for the 2 tests for `Deserialize` as we want to make sure our custom `Visitor` is being used.

components/core/src/service.rs
baumanj: I always prefer the encapsulation of keeping things private because it makes it far easier to refactor the code later and encourages deeper thinking about the interface you want to provide. `pub` by default leads to tighter than necessary coupling and I've yet to see an instance where they're a real boon to usability.

christophermaier: +1 for being a driver of "deeper thinking"

components/core/src/service.rs
baumanj: This may be even more readable:
```rust
        let parts: Vec<_> = bind_str.split(':').collect();
        match parts.as_slice() {
            [name, service_group] => Ok(ServiceBind::new(name, service_group)?),
            _ => Err(Error::InvalidBinding(bind_str.to_string())),
        }
```

components/core/src/service.rs
jamesc: And of course the 4 usages in `sup-protocol` failing the build hid many more in `sup` :) Definitely a better idea to make them private.

components/core/src/service.rs
jamesc: @baumanj The main other user  of `new` is an `Into` trait (which can't fail) on `sup_protocol::types::ServiceBind`.  I've reworked `new` to hot return a `Result` and also consume the `ServiceGroup` passed in via an `into()`.  Would this be surprising for a user?

Would love to talk it through with you

components/core/src/service.rs
```diff
@@ -83,7 +81,89 @@ impl FromStr for BindingMode {
...
+
+impl ServiceBind {
+    pub fn new<S: Into<ServiceGroup>>(name: &str, service_group: S) -> Self {
+        Self {
+            name: name.to_string(),
+            service_group: service_group.into(),
+        }
+    }
+
+    pub fn name(&self) -> &String {
```
baumanj:
```suggestion
    pub fn name(&self) → &str {
```
`&String` is rarely needed since `&str` is more flexible.

components/core/src/service.rs
```diff
@@ -83,7 +81,89 @@ impl FromStr for BindingMode {
...
+    }
+}
+
+impl FromStr for ServiceBind {
+    type Err = Error;
+
+    fn from_str(bind_str: &str) -> result::Result<Self, Self::Err> {
+        match bind_str.split(':').collect::<Vec<_>>().as_slice() {
+            [name, sg_str] => {
+                ServiceGroup::from_str(sg_str).and_then(|sg| Ok(ServiceBind::new(name, sg)))
```
baumanj:
```suggestion
                ServiceGroup::from_str(sg_str)..map(|sg| ServiceBind::new(name, sg))
```
In general, if the body of an `and_then` always returns `Ok(…)` (or `Some(…)` if the receiver is an `Option`), it can be replaced with a simpler `map`.

components/core/src/service.rs
baumanj: I like the change to `new` returning `Self` rather than `Result<Self>`. I'm not sure what the other user of `new` you're referring to is, but I'm happy to chat about it.

In general, if there are several users of `ServiceBind` which want to pass different types that all satisfy `Into<ServiceGroup>`, I think that sounds like a good abstraction. If they're all passing the same type, or there aren't many, I think it's better to let `new` be the simpler form:
```rust
    pub fn new(name: &str, service_group: ServiceGroup) -> Self {
```
and let the caller do the `into` on their side.

components/core/src/service.rs
jamesc: I reworked one of the Deserialize tests to make it less of a serde test and actually check the correct struct fields are being created.

jamesc: @christophermaier Are these tests better now ?

christophermaier: @jamesc Yeah, let's merge it!

PR #119, opened 2019-02-15
Add EnvConfig from habitat as env::Config and implement it for ChannelIdent

baumanj:
See https://github.com/habitat-sh/habitat/issues/6168

There are some compromises here since it's not possible to directly reproduce the previous behavior within the context of the `env::Config` trait. That is, the `HAB_BLDR_CHANNEL` environment variable was used as the primary override for the ChannelIdent default, but if that was not set, it would fall back to trying the legacy environment variable: `HAB_DEPOT_CHANNEL`.

Since `env::Config` uses just one environment variable to provide a `ChannelIdent::configured_value` instance, we can't have the same behavior without putting environment variable overriding logic into `ChannelIdent::default`, which kind of defeats the purpose of the uniform `env::Config` approach.

As such, we instead make a clean break with the legacy approach:
- `ChannelIdent::default` returns the stable channel
- `ChannelIdent::configured_value` returns a `ChannelIdent` instance for the  value in  `HAB_BLDR_CHANNEL` if set and the default value otherwise
- `HAB_DEPOT_CHANNEL` no longer has any special treatment

### DIFF
--- a/components/core/src/env.rs
+++ b/components/core/src/env.rs
@@ -15,6 +15,7 @@
 use std;
 use std::env::VarError;
 use std::ffi::{OsStr, OsString};
+use std::str::FromStr;
 
 /// Fetches the environment variable `key` from the current process, but only it is not empty.
 ///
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

PR #120, opened 2019-02-16
[AtomicWriter] fsync() parent directory after rename

stevendanna:
In the PR that introduced AtomicWriter, we removed this pretty early
in the review process.  Here I've written up various bits of investigation
I've done about whether we need to fsync the directory so we can make
an informed decision about whether to leave it out.

---
To ensure that the rename() has been committed to disk, we must call
fsync() on the containing directory. Without this fsync() call,
powerloss (or VM reset, or crash), could result in data loss. Thus,
this affects the durability of the AtomicWriter but not the atomicity
as far as I know.

## Special casing InvalidInput

Not all filesystems support fsync() on directories. For instance, some
filesystems don't even define the fsync operation on directory
descriptors. In this case Linux returns EINVAL:

https://github.com/torvalds/linux/blob/6f0d349d922ba44e4348a17a78ea51b7135965b1/fs/sync.c#L193-L194

The linux manual page for fsync() says:

> EROFS, EINVAL
>   fd is bound to a special file (e.g., a pipe, FIFO, or socket)
>   which does not support synchronization.

(http://man7.org/linux/man-pages/man2/fsync.2.html)

One such filesystem is vboxsf. While we've had a similar directory
fsync in place for package installations without issues, I opted to
add a special case for EINVAL errors to avoid any unintentional
incompatibilities. We cannot also special case EROFS and EBADF because
those currently get mapped, along with many other errors to
ErrorKind::Other and thus we would be ignoring more than we expected.

## What do other people do

- Postgresql syncs the parent directory and doesn't ignore 
  errors. Note that they include 2 additional fsyncs we don't (one on
  the dest file /before/ the rename and on the dest file /after/ the
  rename); however, it isn't clear to me that those are necessary:

  https://github.com/postgres/postgres/blob/97c39498e5ca9208d3de5a443a2282923619bf91/src/common/file_utils.c#L392-L401

- SQLite syncs the parent directory but ignores the error because of
  systems such as AIX where it isn't supported (the official sqlite
  source is hard to link to):

  https://github.com/mackyle/sqlite/blob/master/src/os_unix.c#L3721-L3740

- MySQL ignores the EINVAL, EBADF, and EROFS:

  https://github.com/mysql/mysql-server/blob/4f1d7cf5fcb11a3f84cff27e37100d7295e7d5ca/mysys/my_sync.cc#L139-L157

- Redis doesn't do this fsync at all:

  https://github.com/antirez/redis/blob/4dc69497f509956d298455492fe5ab98a9738cb6/src/aof.c#L1429-L1444

- RocksDB does the fsync after most dir ops, different call sites
  handle the error differently:

  https://github.com/facebook/rocksdb/blob/8959063c9c98be4c97b6724ef5135aa4a40363a0/util/filename.cc#L380

## Windows

We only do this on Windows because FlushFileBuffers() returns an error
if called on a directory, so we can't call this on Windows. However, I
haven't done research into whether the rename() is guaranteed to be
synced to disk on Windows.

### DIFF
--- a/components/core/src/fs.rs
+++ b/components/core/src/fs.rs
@@ -14,7 +14,7 @@
 
 use dirs;
 use std::env;
-use std::fs as stdfs;
+use std::fs;
 use std::io::{self, Write};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
@@ -397,7 +397,7 @@ impl<'a> SvcDir<'a> {
 
     fn create_dir_all<P: AsRef<Path>>(path: P) -> Result<()> {
         debug!("Creating dir with subdirs: {:?}", &path.as_ref());
-        if let Err(e) = stdfs::create_dir_all(&path) {
+        if let Err(e) = fs::create_dir_all(&path) {
             Err(Error::PermissionFailed(format!(
                 "Can't create {:?}, {}",
                 &path.as_ref(),
@@ -720,10 +720,38 @@ impl AtomicWriter {
             self.tempfile.path().to_string_lossy(),
             &self.dest.to_string_lossy()
         );
-        stdfs::rename(self.tempfile.path(), &self.dest)
-        // NOTE(ssd) 2018-11-01: On some filesystems we should fsync
-        // the parent directory here if we wanted this file to also
-        // provide a durability guarantee.
+        fs::rename(self.tempfile.path(), &self.dest)?;
+
+        #[cfg(unix)]
+        self.sync_parent()?;
+
+        Ok(())
+    }
+
+    /// sync_parent syncs the parent directory. This is required on
+    /// some filesystems to ensure that rename(), create(), and
+    /// unlink() operations have been persisted to disk. sync_parent
+    /// ensures the durability of AtomicWriter but is not required for
+    /// the atomocity guarantee.
+    fn sync_parent(&self) -> io::Result<()> {
+        let parent = parent(&self.dest)?;
+        let f = fs::File::open(parent)?;
+        if let Err(e) = f.sync_all() {
+            // sync_all() calls libc::fsync() which will return EINVAL
+            // if the filesystem does not support calling fsync() on
+            // directories. libc's EINVAL is mapped to InvalidInput.
+            if e.kind() == std::io::ErrorKind::InvalidInput {
+                info!(
+                    "Ignoring InvalidInput from sync_all on {}",
+                    parent.to_string_lossy()
+                );
+                Ok(())
+            } else {
+                Err(e)
+            }
+        } else {
+            Ok(())
+        }
     }
 }
 


thesentinels:

Thank you for contributing!

components/core/src/fs.rs
```diff
@@ -687,6 +687,7 @@ fn parent(p: &Path) -> io::Result<&Path> {
 ///
 pub struct AtomicWriter {
     dest: PathBuf,
+    parent: PathBuf,
```
baumanj: It's a small thing, but I don't think we need to add additional state to the struct here when `parent` can be cheaply computed on-demand. To me, the less state our structs carry around, the purer their methods are, and the easier they are to understand. The alternative of adding a local
```rust
        let parent = parent(&self.dest)?;
```
in sync_parent is also one fewer lines and doesn't change the struct unnecessarily on non-unix platforms.

components/core/src/fs.rs
```diff
@@ -720,10 +722,32 @@ impl AtomicWriter {
...
+        Ok(())
+    }
+
+    // sync_parent syncs the parent directory. This is required on
+    // some filesystems to ensure that rename(), create(), and
+    // unlink() operations have been persisted to disk.
+    fn sync_parent(&self) -> io::Result<()> {
+        if let Err(e) = std::fs::File::open(&self.parent).and_then(|f| f.sync_all()) {
+            match e.kind() {
+                // libc's EINVAL is mapped to InvalidInput
```
baumanj: I think a bit more explanation about why we're ignoring `EINVAL` would be helpful. Do we want to ignore it both on `open` and `sync_all` or just one or the two?

components/core/src/fs.rs
```diff
@@ -720,10 +722,32 @@ impl AtomicWriter {
...
+        self.sync_parent()?;
+
+        Ok(())
+    }
+
+    // sync_parent syncs the parent directory. This is required on
+    // some filesystems to ensure that rename(), create(), and
+    // unlink() operations have been persisted to disk.
+    fn sync_parent(&self) -> io::Result<()> {
+        if let Err(e) = std::fs::File::open(&self.parent).and_then(|f| f.sync_all()) {
```
baumanj:
```suggestion
        if let Err(e) = stdfs::File::open(&self.parent).and_then(|f| f.sync_all()) {
```
It's a little strange we use `stdfs` earlier, but `std::fs` here. I'd also be in favor of eliminating the 

```rust
use std::fs as stdfs;
```
import entirely, and using `std::fs` everywhere, or importing without renaming:
```rust
use std::fs;
```
and using `fs::` everywhere. Regardless of what we do, let's be consistent.

components/core/src/fs.rs
```diff
@@ -720,10 +722,32 @@ impl AtomicWriter {
...
+
+        #[cfg(unix)]
+        self.sync_parent()?;
+
+        Ok(())
+    }
+
+    // sync_parent syncs the parent directory. This is required on
+    // some filesystems to ensure that rename(), create(), and
+    // unlink() operations have been persisted to disk.
```
baumanj: I think it's also worth mentioning that the `sync_parent` isn't necessary for the correctness described in the doc comment for `AtomicWriter` itself, only to provide a durability guarantee.

Also, given that, do we want to make a failure in `sync_parent` a failure of `finish`? What does that improve for us?

Additionally, this would be more valuable as a doc comment: `///`

components/core/src/fs.rs
```diff
@@ -720,10 +722,32 @@ impl AtomicWriter {
...
+    fn sync_parent(&self) -> io::Result<()> {
+        if let Err(e) = std::fs::File::open(&self.parent).and_then(|f| f.sync_all()) {
+            match e.kind() {
```
baumanj: When there's only one value we care about, I find a regular `if/else` to be clearer, and it saves a level of nesting:
```rust
            if e.kind() == std::io::ErrorKind::InvalidInput {
                return Ok(());
            } else {
                return Err(e);
            }
```
vs
```rust
            match e.kind() {
                std::io::ErrorKind::InvalidInput => {
                    return Ok(());
                }
                _ => return Err(e),
            }
```

components/core/src/fs.rs
```diff
@@ -720,10 +722,32 @@ impl AtomicWriter {
...
+                    info!(
+                        "Ignoring InvalidInput from sync_all on {}",
+                        &self.parent.to_string_lossy()
+                    );
+                    return Ok(());
+                }
+                _ => return Err(e),
+            }
+        }
+        Ok(())
```
baumanj: I prefer not to have these kinds of "fall-through" return values because if a previous early return gets removed, we can end up returning `Ok(())` somewhat accidentally. I prefer the structure
```rust
        if let Err(e) = … {
            …
        } else {
            Ok(())
        }
```
because it additionally allows the early returns in the `if let Err` half of the conditional to be converted to regular expression-returns rather than early-out return statements and lets the compiler help us recognize if we introduce some kinds of logic errors with respect to our return value.

stevendanna: Cool, the former is our prevaling go style but I like your proposal better as well in general

stevendanna:
> Take whatever you like and let me know when you feel it's merge-ready.

Thanks for the review. I mostly typed this up just to record the research around this.  I can go either way about whether we do this fsync or not. 

As a heads up, probably won't get back to this for a day or two.

components/core/src/fs.rs
stevendanna: Good catch, I think we only care to special case EINVAL from sync_all (fsync).  Open can also return EINVAL in these cases:

```
       EINVAL The filesystem does not support the O_DIRECT flag.  See NOTES for more information.

       EINVAL Invalid value in flags.

       EINVAL O_TMPFILE was specified in flags, but neither O_WRONLY nor O_RDWR was specified.

       EINVAL O_CREAT was specified in flags and the final component ("basename") of the new file's pathname is invalid (e.g., it contains characters not permitted by the underlying filesystem).
```

All of those seem rather unlikely given my reading of libstd currently, but better to be safe here and only special case the sync_all call.

stevendanna:
I believe I've addressed the various style concerns here and I also limited the error special-case to the sync_all.

baumanj:
This is such awesome, thoroughly-researched work. Thanks, @stevendanna !

PR #126, opened 2019-03-02
Remove output module to be transferred to habitat repo

mwrock:
This replaces the use of `ansi_term` with `termcolor` for coloring console output.Because of the stateful nature of coloring with termcolor (thanks to the need to call Windows terminal APIs), we need to give fore control to `StructuredOutput` in handling how it streams its output.

Note that this removes the `output_format!` macro which was only used by the launcher. The launcher will now create its own `StructuredOutput` and call either `println` or `eprintln`.

We also give `StructuredOutput` a `ColorSpec` for its content which is used by the supervisor ctrl gateway.

see https://github.com/habitat-sh/habitat/pull/6256 for corresponding habitat repo PR


thesentinels:

Thank you for contributing!

components/core/src/output.rs
```diff
@@ -101,6 +102,8 @@ pub struct StructuredOutput<'a> {
     pub color: Option<bool>,
     /// Whether or not to render as structured JSON logging output.
     pub json: Option<bool>,
+    /// Color and styling to use for content.
+    pub color_spec: ColorSpec,
```
baumanj: Does this need to be public? This is a new field, so I think we can do better by creating an more limited interface.

While we're at it `color` and `json` can have their `pub`s removed, as nobody's using them. And at that point, we'd just need to add an interface for `verbose` and we'd have good encapsulation.

baumanj: While we're at it, the `color` field seems entirely unused outside tests. Can we eliminate it? The same goes for the `json` field. It seems weirdly awkward to have these options that complicate the structure, but aren't really used.

components/core/src/output.rs
```diff
@@ -123,6 +126,102 @@ impl<'a> StructuredOutput<'a> {
...
+        buffer.write_all(b"\n")?;
+        buffer.flush()?;
+        writer.print(&buffer)
+    }
+
+    fn color_choice(&self) -> ColorChoice {
+        match self.color.unwrap_or_else(is_color) {
+            true => ColorChoice::Auto,
+            false => ColorChoice::Never,
+        }
```
baumanj: I think a regular `if` is more concise:
```rust
        if self.color.unwrap_or_else(is_color) {
            ColorChoice::Auto
        } else {
            ColorChoice::Never,
        }
```
and if we can eliminate the `color` field, even more so:
```rust
        if is_color() {
            ColorChoice::Auto
        } else {
            ColorChoice::Never
        }
```

components/core/Cargo.toml
```diff
@@ -11,7 +11,6 @@ base64 = "*"
 gcc = "0.3"
 
 [dependencies]
-ansi_term = "*"
```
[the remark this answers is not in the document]
mwrock: should be good now

components/core/src/fs.rs
```diff
@@ -733,6 +733,7 @@ impl AtomicWriter {
     /// unlink() operations have been persisted to disk. sync_parent
     /// ensures the durability of AtomicWriter but is not required for
     /// the atomocity guarantee.
+    #[cfg(unix)]
```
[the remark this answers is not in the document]
mwrock: no it builds but it emits a warning `sync_parent` is not used

components/core/src/output.rs
mwrock: if we eliminate them I think testing json vs color vs no color output will not be reliable in a multithreaded testing context. I found these tests helpful in developing this.

components/core/src/output.rs
mwrock: I think you caught this as I was making some changes. I added in a check for json since json formatting will suppress color and did go for an `if` over `match` approach.

mwrock:
related to https://github.com/habitat-sh/habitat/issues/2122

components/core/src/output.rs
mwrock: I was just thinking of this and I think what we want to do is keep the private `color`, `json` and `verbose` fields, add them to the constructor and do not call into the outside `AtomicBool` functions at all from the struct. The struct should be entirely decoupled from its environment and lett the callers, check `is_json`, `is_color`, et al and pass the correct value to the constructor. Does that sound good to you @baumanj ?

components/core/src/output.rs
baumanj: Got it. Since there are still only two outcomes, what about
```rust
        if self.color.unwrap_or_else(is_color) && !self.json.unwrap_or_else(is_json) {
            ColorChoice::Auto
        } else {
            ColorChoice::Never
        }
```
?
It would read a lot cleaner if we could get rid of the `Option`s:
```rust
        if is_color() && !is_json() {
            ColorChoice::Auto
        } else {
            ColorChoice::Never
        }
```
Come to think of it, why do we allow the static variables `JSON` and `NO_COLOR` to be set to an inconsistent combination?

components/core/src/output.rs
```diff
@@ -154,59 +270,13 @@ impl<'a> Serialize for StructuredOutput<'a> {
...
-                    White.bold().paint(self.logkey),
-                    self.content
-                )
-            } else {
-                write!(f, "{}({}): {}", self.preamble, self.logkey, self.content)
-            }
+        let bufwtr = BufferWriter::stdout(self.color_choice());
+        let mut buffer = bufwtr.buffer();
+        match self.format(&mut buffer) {
+            Ok(_) => f.write_str(&String::from_utf8_lossy(buffer.as_slice())),
```
baumanj: Are we certain that `buffer` will contain valid utf8? If so, why not
```suggestion
            Ok(_) => f.write_str(std::str::from_utf8(buffer.as_slice()).expect("termcolor buffer valid utf8")),
```
And if we're not, how might we be getting invalid utf8 in there?

components/core/src/output.rs
baumanj: Yeah, that sounds excellent!

components/core/src/output.rs
mwrock: I was actually thinking earlier of adding a `OutputFormat` enum that can be `Color`, `NoColor`, or `JSON`. That would remove some confusion I think and also simplify this conditional.

mwrock:
Yeah I will move this into `habitat/components/common` along with https://github.com/habitat-sh/habitat/pull/6256 which should change this PR into just a deletion.

components/core/src/output.rs
baumanj: I like it. If you go the proper `enum` route, I think the `static`s will need to be changed a `Mutex` containing the enum with `lazy_static!`.

You could also do an `AtomicUsize` or something, but that would require casts or bitfields, and I don't think it's worth it.

components/win-users/Cargo.toml
```diff
@@ -14,5 +14,5 @@ gcc = "0.3"
 log = "*"
 
 [target.'cfg(windows)'.dependencies]
-widestring = "*"
+widestring = "0.3.0"
```
baumanj: Why are we pinning to a particular version? There may be a good reason, but since it's contrary to our normal practice, I think it should be commented with a link to an issue (if appropriate).

mwrock: comment added

mwrock:
fixes https://github.com/habitat-sh/core/issues/128

PR #130, opened 2019-03-05
Adds BUILD_DEPS and BUILD_TDEPS to available metadata types

eeyun:
Some necessary additions for making build deps integration to the build graph.

### DIFF
--- a/components/core/src/package/archive.rs
+++ b/components/core/src/package/archive.rs
@@ -210,7 +210,6 @@ impl PackageArchive {
     ///
     /// # Failures
     ///
-    /// * If a `DEPS` metafile is not found in the archive
     /// * If the archive cannot be read
     /// * If the archive cannot be verified
     pub fn deps(&mut self) -> Result<Vec<PackageIdent>> {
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/core/src/package/archive.rs
```diff
@@ -229,6 +229,30 @@ impl PackageArchive {
         self.read_deps(MetaFile::TDeps)
     }
 
+    /// Returns a list of package identifiers representing the buildtime package dependencies for
+    /// this archive.
+    ///
+    /// # Failures
+    ///
+    /// * If a `BUILD_DEPS` metafile is not found in the archive
```
baumanj: So, if there aren't actually any build deps (or similarly tdeps), there *will* be a file, but it will be empty?

eeyun: Nope, if it works the way the other metadata files work, if there are no deps of the type then the file wont exist.

baumanj: In that case the doc comment here is incorrect, right? Or maybe I'm misunderstanding it.

christophermaier: This documentation comment (and the other similar ones in this module) _are_ incorrect. The `read_deps` function will return an empty vector if there's no file.

PR #131, opened 2019-03-05
Add purge_templated_content method

christophermaier:
This will eventually be used by the Supervisor to remove rendered content from older package versions which have been removed in current versions.

See habitat-sh/habitat#4925 for further background and context.

### DIFF
--- a/components/core/src/fs.rs
+++ b/components/core/src/fs.rs
@@ -339,6 +339,48 @@ impl<'a> SvcDir<'a> {
         Ok(())
     }
 
+    /// Remove all templated content (hooks and configuration) from a
+    /// service directory.
+    ///
+    /// Useful for removing rendered files that may be from older
+    /// versions of a service that have been removed from the current
+    /// version.
+    pub fn purge_templated_content(&self) -> Result<()> {
+        for dir_path in &[svc_config_path(&self.service_name),
+                          svc_hooks_path(&self.service_name)]
+        {
+            debug!("Purging any old templated content from {}",
+                   dir_path.display());
+            Self::purge_directory_content(dir_path)?;
+        }
+        Ok(())
+    }
+
+    /// Utility function that removes all files in `root`.
+    fn purge_directory_content<P>(root: P) -> Result<()>
+        where P: AsRef<Path>
+    {
+        for entry in fs::read_dir(root.as_ref())? {
+            let entry = entry?;
+            if let Ok(file_type) = entry.file_type() {
+                if file_type.is_file() {
+                    debug!("Purging file {:?}", entry.path().display());
+                    fs::remove_file(entry.path())?;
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/core/src/fs.rs
```diff
@@ -350,6 +352,54 @@ impl<'a> SvcDir<'a> {
...
+            );
+            Self::purge_directory_content(dir_path)?;
+        }
+        Ok(())
+    }
+
+    /// Utility function that removes all files in `root`.
+    fn purge_directory_content<P>(root: P) -> Result<()>
+    where
+        P: AsRef<Path>,
```
baumanj: Especially for a private function, I prefer to keep things simpler unless generics are a big ergonomics win:
```rust
    fn purge_directory_content(root: &Path) -> Result<()> {
```
In this case, I think it's a net simplification. The changes required for this are:
```diff
-        for entry in fs::read_dir(root.as_ref())? {
+        for entry in fs::read_dir(root)? {
```
```diff
-            SvcDir::purge_directory_content(&root).expect("Couldn't purge!");
+            SvcDir::purge_directory_content(root.path()).expect("Couldn't purge!");
```

components/core/src/fs.rs
```diff
@@ -350,6 +352,54 @@ impl<'a> SvcDir<'a> {
...
+                } else if file_type.is_symlink() {
+                    debug!("Purging symlink {:?}", entry.path().display());
+                    fs::remove_file(entry.path())?;
+                }
+            } else {
+                warn!(
+                    "Not purging {:?}; could not determine file type",
+                    entry.path().display()
+                );
+            }
```
baumanj: There's a fair bit of duplication here which makes it a bit harder to see the code. Also, it would be nice to print the error message along with our `warn` message. Finally, there's no need to use `{}` for `display()`, and as long as we're using debug messages, we can reduce code by using `{:?}` on `entry` rather than `{}` on `entry.path().display()`. All together:
```rust
            let file_type_result = entry.file_type();
            debug!("Attempting purge of {:?} {:?}", file_type_result, entry);
            match file_type_result {
                Ok(ft) if ft.is_file() || ft.is_symlink() => fs::remove_file(entry.path())?,
                Ok(ft) if ft.is_dir() => fs::remove_dir_all(entry.path())?,
                Ok(ft) => debug!("Nothing to do for {:?}", ft),
                Err(e) => {
                    warn!(
                        "Not purging {}; could not determine file type: {}",
                        entry.path().display(),
                        e
                    );
                }
            }
```

christophermaier:
@baumanj Aaaaah, yeah... I was meaning to fix those. The code looked very different in an earlier incarnation; incomplete refactoring strikes again 😦 

I'll get those bits fixed up.

PR #140, opened 2019-03-19
Raise client connection timeout

chefsalim:
We need to raise the default timeout for the API client in order to accommodate large file uploads to Builder.  This (along with adjustment to the ELB idle timeout) resolves https://github.com/habitat-sh/builder/issues/961
 
Signed-off-by: Salim Alam <salam@chef.io>

tenor-76384458

### DIFF
--- a/components/http-client/src/api_client.rs
+++ b/components/http-client/src/api_client.rs
@@ -45,7 +45,7 @@ use crate::{error::{Error,
             ssl};
 
 // Read and write TCP socket timeout for Hyper/HTTP client calls.
-const CLIENT_SOCKET_RW_TIMEOUT_SEC: u64 = 120;
+const CLIENT_SOCKET_RW_TIMEOUT_SEC: u64 = 300;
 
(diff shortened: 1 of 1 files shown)


thesentinels:

Thank you for contributing!

components/http-client/src/api_client.rs
```diff
@@ -45,7 +45,7 @@ use crate::{error::{Error,
             ssl};
 
 // Read and write TCP socket timeout for Hyper/HTTP client calls.
-const CLIENT_SOCKET_RW_TIMEOUT_SEC: u64 = 120;
+const CLIENT_SOCKET_RW_TIMEOUT_SEC: u64 = 300;
```
christophermaier: Can you use the [Config](https://github.com/habitat-sh/core/blob/83d305e4acfa83f1b8925d9ecea58955a36a448f/components/core/src/env.rs#L91) trait to provide an override for folks?

chefsalim: I'm happy to do that at a later time if/when we touch this code again. Currently it does have an env override so we have the desired functionality.

PR #141, opened 2019-03-19
Correctly run clippy in CI

baumanj:
Clippy leverages `cargo check`, so running it after other tests may cause some lints to not be triggered. To increase effectiveness (and move fatal errors toward the bottom of the CI output), run clippy first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,9 +49,9 @@ matrix:
         - ./support/ci/compile_zmq.sh
         - source ./support/ci/rust_env.sh
       script:
+        - make lint
         - ./support/ci/rust_tests.sh
         - ./support/ci/rustfmt.sh
-        - make lint
 
 notifications:
   webhooks:


thesentinels:

Thank you for contributing!

raskchanky:
@christophermaier I'm guessing the issue is that running `cargo check` _after_ doing `cargo build` or `cargo test` won't actually build anything, since cargo is smart enough to detect that no code was changed.

christophermaier:
Hrmm... interesting 🤔 

Does that suggest that we'd need to run `cargo clean` before running clippy? That seems a tad extreme.

baumanj:
> Does that suggest that we'd need to run `cargo clean` before running clippy? That seems a tad extreme.

In this case, I think it suffices to move it to the first position so that we don't have any previous artifacts, but it is indeed possible for clippy to miss things on account of the erroneous code not being reconsidered at all.

See https://github.com/rust-lang/rust-clippy/issues/2604
and https://github.com/rust-lang/cargo/pull/6664

christophermaier:
@raskchanky @baumanj Cool, thanks for all that. 👍

PR #142, opened 2019-03-22
Clarify signal behavior

christophermaier:
Our `Signal`-related code makes it appear that there is a common contract across Linux and Windows. In reality, there is _not_, and making it appear otherwise can create misconceptions and complications elsewhere.

This PR rearranges and streamlines code to make the cross-platform differences clear, with the goal of making future refactorings easier.


thesentinels:

Thank you for contributing!

components/core/src/os/signals/unix.rs
```diff
@@ -74,13 +73,38 @@ pub fn check_for_signal() -> Option<SignalEvent> {
...
+/// some kind of event
+fn from_signal_code(code: SignalCode) -> Option<Signal> {
+    // only 3 signals are explicitly acted upon... everything else
+    // is passed through to the supervisor
+    //
+    // If it gets a HUP, it restarts the Supervisor
+    // process... anything else will pretty much kill the
+    // supervisor outright (I think)
+    //
+    // Also, these are things that get sent to the launcher.
```
baumanj: There's nothing consumer-specific about the implementation here; I'd remove those parts from the comments.

components/core/src/os/process/windows.rs
```diff
@@ -28,21 +28,13 @@ use winapi::{shared::minwindef::{DWORD,
                           PROCESS_QUERY_LIMITED_INFORMATION,
                           PROCESS_TERMINATE}}};
 
-use super::{OsSignal,
-            Signal};
+use super::Signal;
```
mwrock: This needs to go too since it got zapped and id no longer being used here.

christophermaier: tenor-29596731

components/core/src/os/signals/unix.rs
christophermaier: Ah, sorry... those were some notes to myself as I was trying to figure out what was what. I'll revise or remove.

components/core/src/os/signals/windows.rs
```diff
@@ -14,29 +14,18 @@
...
-    if CAUGHT.load(Ordering::SeqCst) {
-        // clear out the signal so we don't sent it repeatedly
-        CAUGHT.store(false, Ordering::SeqCst);
-        Some(SignalEvent::Shutdown)
-    } else {
-        None
-    }
-}
+/// Returns `true` if we have received a signal to shut down.
+pub fn check_for_shutdown() -> bool { CAUGHT.compare_and_swap(true, false, Ordering::SeqCst) }
```
baumanj: Since this is essentially identical to the linux implementation, is there any way to share the code rather than duplicate it?

christophermaier: That occurred to me, too. I think so, but it'll require a bit more shuffling things around.

christophermaier: @baumanj see what you think about https://github.com/habitat-sh/core/pull/142/commits/618a8532f7ae3d43f90c5ec5b98f85c0439eca86

baumanj: I like it!

PR #144, opened 2019-03-27
Adds more build_dep and build_tdep functionality

eeyun:
Not extracting the build_dep and build_tdep data into the hashmap meant not being able to actually utilize this information in any meaningful way.

### DIFF
--- a/components/core/src/package/archive.rs
+++ b/components/core/src/package/archive.rs
@@ -76,6 +76,22 @@ lazy_static::lazy_static! {
             .unwrap(),
         );
         map.insert(
+            MetaFile::BuildDeps,
+            Regex::new(&format!(
+                r"^/?hab/pkgs/([^/]+)/([^/]+)/([^/]+)/([^/]+)/{}$",
+                MetaFile::BuildDeps,
+            ))
+            .unwrap(),
+        );
+        map.insert(
+            MetaFile::BuildTDeps,
+            Regex::new(&format!(
+                r"^/?hab/pkgs/([^/]+)/([^/]+)/([^/]+)/([^/]+)/{}$",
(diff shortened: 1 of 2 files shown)


thesentinels:

Thank you for contributing!

components/core/src/package/archive.rs
```diff
@@ -75,6 +75,22 @@ lazy_static::lazy_static! {
...
+            Regex::new(&format!(
+                r"^/?hab/pkgs/([^/]+)/([^/]+)/([^/]+)/([^/]+)/{}$",
+                MetaFile::BuildDeps,
+            ))
+            .unwrap(),
+        );
+        map.insert(
+            MetaFile::BuildTDeps,
+            Regex::new(&format!(
+                r"^/?hab/pkgs/([^/]+)/([^/]+)/([^/]+)/([^/]+)/{}$",
```
baumanj: As long as we're here, should we abstract out the `r"^/?hab/pkgs/([^/]+)/([^/]+)/([^/]+)/([^/]+)/{}$"` regex that we duplicate 10 times?

eeyun: In the future that would be good, but it's out of scope for this change. I can add a TODO comment to the file if you'd like?

baumanj: I don't think TODO comments are helpful. My order of preference would be:

1. A separate PR directly after this one to de-duplicate
1. File an issue to de-duplicate
1. Do nothing

eeyun: I will open an issue for now, I don't want to deal with more context switching on this currently.

eeyun:
RE: regex abstraction - https://github.com/habitat-sh/core/issues/145

PR #148, opened 2019-04-12
Bump to Rust 1.34.0

christophermaier:


components/core/src/crypto/keys/box_key_pair.rs
```diff
@@ -497,13 +497,12 @@ mod test {
         assert_eq!(pair.public, None);
         match pair.public() {
             Ok(_) => panic!("Empty pair should not have a public key"),
-            Err(Error::CryptoError(_)) => assert!(true),
+            Err(Error::CryptoError(_)) => { /* OK */ }
```
baumanj: We could simplify this a bit:
```rust
if let Err(Error::CryptoError(_)) = pair.public() {
    // OK
} else {
    panic!("Expected Error::CryptoError");
}
```
It reduces the information that's produced in the case of failing with a _different_ error, but I'm ok with that tradeoff. I'll leave it up to you.

components/core/src/crypto/keys/box_key_pair.rs
```diff
@@ -497,13 +497,12 @@ mod test {
...
             e => panic!("Unexpected error: {:?}", e),
         }
         assert_eq!(pair.secret, None);
-        match pair.secret() {
-            Ok(_) => panic!("Empty pair should not have a secret key"),
-            Err(_) => assert!(true),
+        if pair.secret().is_ok() {
+            panic!("Empty pair should not have a secret key");
```
baumanj: Or
```rust
assert!(pair.secret().is_err(), "Empty pair should not have a secret key");
```
Similarly elsewhere

components/core/src/package/archive.rs
```diff
@@ -411,7 +410,7 @@ impl PackageArchive {
 
         match self.read_metadata(file) {
             Ok(Some(body)) => {
-                let ids: Vec<String> = body.lines().map(|d| d.to_string()).collect();
+                let ids: Vec<String> = body.lines().map(ToString::to_string).collect();
```
baumanj: I think
```suggestion
                let ids: Vec<String> = body.lines().map(str::to_string).collect();
```
or
```suggestion
                let ids: Vec<String> = body.lines().map(str::to_owned).collect();
```
are more illustrative. Unfortunately clippy makes the suggestion which is most broadly applicable, but when our type is know beyond just the trait bound, I find it makes it clearer what's going on.

components/core/src/package/ident.rs
```diff
@@ -528,9 +526,8 @@ mod tests {
                                   "banana".to_string(),
                                   Some("1.0.0".to_string()),
                                   Some("20150521131555".to_string()));
-        match a.partial_cmp(&b) {
-            Some(_) => panic!("We tried to return an order"),
-            None => assert!(true),
+        if a.partial_cmp(&b).is_some() {
+            panic!("We tried to return an order");
```
baumanj: Why not
```rust
        assert!(a.partial_cmp(&b).is_none(), "We tried to return an order");
```
?

components/core/src/package/metadata.rs
```diff
@@ -62,7 +61,7 @@ impl FromStr for Bind {
         };
         let exports = match parts.next() {
             None => return Err(Error::MetaFileBadBind),
-            Some(exports) => exports.split(' ').map(|t| t.to_string()).collect(),
+            Some(exports) => exports.split(' ').map(ToString::to_string).collect(),
```
baumanj:
```suggestion
            Some(exports) => exports.split(' ').map(str::to_string).collect(),
```

baumanj (approved):
Some little suggestions, but all optional

components/core/src/crypto/keys/box_key_pair.rs
christophermaier: Yup, that's much better 😄

raskchanky (approved):

